PR #10, opened 2015-06-17
Correct tracking and display on Multisite

swissspidy:
What this patch does:
- Track the correct absolute path on Multisite. For example, when you visit `http://example.com/subsite/`, don't track `/subsite/subsite/`, but only `/subsite/`
- In the stats widget, use `home_url` to build the right URL again for this site.


glueckpress:
Reviewed by @danielhuesken, thx!

PR #25, opened 2016-09-09
Update README.md

timse201:


patrickrobrecht:
I think the German paragraph should be deleted because the German page should in the repository https://de.wordpress.org/plugins/statify/ (to be translated on https://de.wordpress.org/plugins/statify/ - currently not stable)

Zodiac1978 (approved):

bueltge:
Thanks :heart: for contributing! 
@patrickrobrecht right, good point and a result of older times without different languages of the site. It is now a topic for the next release on wp.org.

PR #26, opened 2016-09-09
Delete the russian translation

timse201:
after pushing these 4 translations we can delete it

[WordPress Translation](https://translate.wordpress.org/projects/wp-plugins/statify/dev/ru/default?filters%5Btranslated%5D=yes&filters%5Bstatus%5D=waiting)


patrickrobrecht:
As the Russian translation is the last one, we can also simplify loading the textdomain to `load_plugin_textdomain( 'statify' )`, the reference to the lang directory isn't needed anymore.

swissspidy:
+1 for considering simplifying the textdomain loading in a separate PR afterwards.

This PR only can be merged as soon as the translation reaches 100% on translate.wordpress.org.

patrickrobrecht:
The missing strings have been approved and the Russian language pack has been generated. So this PR can be merged now.

Issue #28, opened 2016-09-23
Bounding box is not moving in dashboard widget


sesom42:
The bounding box does not follow the mouse. On entering the widget the box is displayed on a static position. Tested with FF 49.0 and Chrome 53.0.

![statify box problem ong](https://cloud.githubusercontent.com/assets/8649051/18778032/a04d43e8-8172-11e6-91ad-d7ad366b2d4c.png)

websupporter:
It's not just the box, sometimes the text is not in the right direction:
![js-error](https://cloud.githubusercontent.com/assets/6458412/18780369/de5a66ee-8185-11e6-8de3-cf70406f62eb.jpg)

I suspect right now the helper function `popup()`. What I've noticed so far is the `dx` jumps quite a lot. When I get negative values, my text is in the wrong place. But haven't figured out yet, whats going on.

Crossreferencing: https://wordpress.org/support/topic/statistics-displaying-incorrectly/ 
It doesn't look like a conflict between antispam bee and statify, but more a unique js error in the raphael-helper.

websupporter:
Hi @sesom42, I've tried to reproduce your specific problem. I did encounter some, but can't reproduce the box not moving to the supposed position. I am testing on FF49 as well. Can you give a more detailed path to how to reproduce it? For example, the problem I encountered above, I can reproduce it when highlighting the same point over and over again. In this case it jumps (not on a regular base though) the way you can see in the screenshot.

When I simply enter the widget with the mouse or try a bit I was not able to reproduce the problem showing in your screenshot.

Thank you.

sesom42:
I have tested with all plugins disabled and then the box is displayed correctly. After enabling my plugins step by step I have found the plugin that triggers the issue: Antispam Bee. If the dashboard widget in Antispam Bee is enabled then the box in Statify does not follow the mouse cursor.

websupporter:
Thanks @sesom42 for reporting. I've made a pull request which would solve this problem.

PR #29, opened 2016-09-28
Unify Libs

websupporter:
Would solve https://github.com/pluginkollektiv/statify/issues/28

Haha, I couldn't reproduce it, because I did already unify the libs in my local install while I was testing the problem...

This solves the problem of the popup not moving to its position. The problem I have reported in this issue is a seperated one. It appears without unification and with unification and needs seperate attention.


patrickrobrecht:
Checked that this pull request really solves the issue.

PR #31, opened 2016-10-22
Documentation to README

patrickrobrecht:
This PR would move all the documentation from the Wiki to the README file such that it can be translated via translate.wordpress.org as the general description etc.


Zodiac1978 (approved):

bueltge:
Fixed #35 
thanks for contributing @patrickrobrecht

PR #36, opened 2017-01-08
Tabnapping prevention; Link to Support, FAQ added; PayPal, Wiki translateable; Renamed PayPal to Donate

timse201:


bueltge (commented):
Please sanitize the strings for translations, like ` __( ` to `esc_html__( `.
Also in my opinion we should remove the target blank attribute, not really helpful for the users. but other longer topics.

bueltge:
Thanks a lot!

Issue #16, opened 2015-08-17
WordPress' coding standards


ghost:
Hello,

Parallel to this project, I've been using a fork containing some adjustments to match [WordPress' coding standards](https://github.com/pluginkollektiv/statify).

Is this something this project might consider?

swissspidy:
> Is this something this project might consider?

Absolutely. Since we have quite a few plugins to work on, we need to find a great solution to establish coding standards across all projects.

Grunt tasks (using [dedicated PHP_CodeSniffer rules](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards)) are a great way to always enforce the current coding standards. These can also be used for unit tests and CI services like Travis.

A sample setup can be seen in my [oEmbed feature plugin](https://github.com/swissspidy/oEmbed-API). As you can see, coding standards are not the only problem. We should also strive for simple code with great documentation and without duplicated logic.

ghost:
<blockquote>Grunt tasks (using dedicated PHP_CodeSniffer rules) are a great way to always enforce the current coding standards. These can also be used for unit tests and CI services like Travis.</blockquote>

I'll have to update my fork its latest release, but regarding the WordPress coding standards, I could create a pull request within a couple of days.

Unit test is still something I have almost no experience with.

swissspidy:
I suggest doing small steps here. Although we're always happy about new pull requests, we need to discuss the ideal approach internally first. Statify isn't the only plugin in need for this :)

ghost:
Understandable :)

bueltge:
@corvannoorloos I switch all files to the first idea of codex. Maybe you will help also yet? We need all. Currently I miss also a dev stylesheet to write on this, the minified version should a result of grunt (maybe gulp). 
Also for a codex is a readme.txt necessary, currently missed.

swissspidy:
@bueltge WordPress.org handles readme.md files just fine

bueltge:
Yes, i saw it. But the [official documentation](https://wordpress.org/plugins/about/) would a txt, custom format. Also we have the chance to get a difference between github and wp.org views. I think the effort is small, but the result is big.

PR #38, opened 2017-01-24
readme changes

timse201:
readme.txt:
documentation link
1 more faq entry
standardized (think we should have 1 standard in all these repos)
code in boxes
settings as a list and not Italic

readme.md
Donate + FAQ links

#16


bueltge (approved):

bueltge:
Thanks a lot, really helpful.

PR #41, opened 2017-01-30
update donation links

timse201:
(as discussed on slack)


bueltge (approved):

bueltge:
In my opinion is the donation link not necessary. Where is the result of the donation? What is with this result? I think we build open source solutions, helpful for other and a part of learning for us on the maintenance and development. I think it is much more helpful that we cry for helpful hands ;)

timse201:
True
I agree I do it because of fun and learning.

PR #51, opened 2017-08-13
Update compatibility info to WP 4.7+

patrickrobrecht:
- Updating the minimum WP version to 4.7 fixes a compatibility issue as we use the second parameter of `wp_parse_url` which is [not available before WP 4.7](https://developer.wordpress.org/reference/functions/wp_parse_url/#changelog) ([bug report](https://wordpress.org/support/topic/problem-beste-inhalte/)).
- In WP 4.7+ the fallback for `get_sites()` is not necessary anymore.


bueltge:
+1

bueltge (approved):

PR #53, opened 2017-11-27
Update Raphaël to 2.2.7 (fix #52)

### LINKED ISSUE #52: Raphaël does not work when CSP disallows 'unsafe-eval'
The version of Raphaël used in this plugin does not work on pages with a Content Security Policy disallowing `unsafe-eval` for [`script-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). This was fixed in Raphaël version [2.1.2](https://github.com/DmitryBaranovskiy/raphael/blob/master/history.md#212--2015-08-26).

stklcode:
As the headline indicates, the Raphaël JS libaray has been updates to the latest version 2.2.7 straight from the GItHub repo.

As far as I can tell this does not break anything, but fixes #52. As long as we don't come around with an alternative as discussed in #1, this should be the fastest fix. Someone should probably double-check this :wink:


bueltge:
Really great! :+1:

bueltge (approved):

PR #59, opened 2017-12-02
Do not set code 204 on JS snippet (fix #58)

### LINKED ISSUE #58: JavaScript tracking doesn’t work on Safari (and maybe Chrome?) with X-Content-Type: nosniff
I generally enable `X-Content-Type: nosniff` on all of my websites because it’s an easy way to preserve Cross-Site-Scripting. Unfortunately, after enabling Cachify and thus also enabling the JavaScript tracking method in Statify Safari is unable to be tracked.

In the console of Safari the following error message appears:
```
Did not load script at 'https://kittpress.com/?statify_referrer=&statify_target=%2F' because non script MIME types are not allowed when 'X-Content-Type: nosniff' is given.
```

You can test it by yourself if you visit https://kittpress.com.

By using Chrome there isn’t even any entry in the network tab that refers to the Statify JavaScript caching and as far as I interpret my tracking statistics in the Dashboard there isn’t any tracking for Chrome users at all.

Maybe it could be enough to properly send a valid mime type?

stklcode:
Content-Type header might be omitted if status code is 204 (no content) which effectively breaks JS tracking in 'nosniff' environments, because the "script" inclusion is prohibited by modern browsers.

Makes sense somehow to set 204 on empty respnse, but it also makes sense to ignore the type of "nothing"... Simply removing the explicit status code and leaving the default 200 solves the issue.

### DIFF
--- a/inc/statify_frontend.class.php
+++ b/inc/statify_frontend.class.php
@@ -224,7 +224,7 @@ class Statify_Frontend extends Statify {
 
 		if ( $is_snippet ) {
 			nocache_headers();
-			header( 'Content-type: text/javascript', true, 204 );
+			header( 'Content-type: text/javascript', true );
 			exit;
 		}
 


stklcode:
There is another possible solution for this problem.

At the moment the JS snippet creates a `<script>`-tag with the tracking endpoint URL as `src `. Alternatively it could issue a GET request directly via XHR (then the 204 would be fine).

There are pros and cons for both versions. We shouldn't expect to run into CORS issues with the second approach, as the request is fired to a relative URL via GET which matches far most setups and inline JS has to be evaulated with both methods.

Open for discussion, if this solution would be preferred, I could change the PR to this branch: stklcode/statify@f5f69c4

bueltge (approved):

chesio:
I like the XHR approach more, it feels more clean. I don't see any cons here: as you mentioned, it's not a cross-origin request, so CORS policy should not come into play at all. And CSP must allow inline script execution in both cases.

websupporter:
To be honest, I have no opinion on which way to go. If removing `204` solves the problem, why not just going down this road.

stklcode:
Oops, forgot about this PR while updating my fork, didn't intend to close it... Shame on me :smile:

patrickrobrecht:
I agree with @chesio that the XHR approach looks cleaner.

patrickrobrecht:
Closed in favor of #68

PR #64, opened 2018-02-02
fix license declaration in composer.json

websupporter:
License "GPL-3.0" is a deprecated SPDX license identifier, use "GPL-3.0-only" or "GPL-3.0-or-later" instead

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "pluginkollektiv/statify",
   "description": "Compact, easy-to-use and privacy-compliant stats plugin for WordPress.",
-  "license": "GPL-3.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "pluginkollektiv/statify",
   "description": "Compact, easy-to-use and privacy-compliant stats plugin for WordPress.",
-  "license": "GPL-3.0",
+  "license": "GPL-3.0-only",
```
stklcode, line 4: Maybe merged too fast... PHP file says "GPLv3 or later".

bueltge:
For documentation, this keys are valid https://spdx.org/licenses/

Issue #66, opened 2018-02-18
Chart Legibility


manumeter:
Hi everyone,

I really appreciate your great and simple WP plug-in, thank you very much!

My issue is that I want to keep the (anonymized) statistics for a couple of years. However, the legibility of the chart gets almost lost after about half a year of statistics and it will even get worse over time.

![screenshot from 2018-02-18 15-27-04](https://user-images.githubusercontent.com/4171359/36352982-a5924d02-14c0-11e8-88f3-fa9a59143097.png)

It would be great if you could handle this by either:
 - thinning the data out to e.g. only one point per week/month with the sum of visits or
 - make the chart wider and scrollable or
 - make the chart configurable to only display a subset (e.g. the last four weeks) by still keeping all data

Thank you very much for considering enhancing your great plug-in to also match my use case completely.

Best regards!

stklcode:
> thinning the data out to e.g. only one point per week/month with the sum of visits

Sound like the best fitting idea from my point of view. Determine the maximum viable number of entries for the chart (e.g. 90?) and aggregate to week-level after that month-level after 90 weeks. Don't think a coarser scale is required. Maybe adding some JS functionality to allow scrolling (just brainstorming).

Possible argument against that would be the amount of data. With a frequency as shown above it might be suitable to store raw data and aggregate on-the-fly, but this could easily end up in millions of entries over time. Probably not that much of an issue on modern server systems, aggregating the data on a daily/monthly/... basis into a separate data structure (e.g. via hook in cleanup-phase or cron job) and adding an extended long-term chart might be considered a better solution. Sounds more like an extension plugin though.


And to complete my thoughts, what about just clipping the chart? Allow configuring storage and view timespan separately to keep the everyday dashboard clean and only look into archive data when needed?

patrickrobrecht:
This issue is partly fixed in #67 , but for installations keeping data for long time periods (> 2 years), the chart on the dashboard still does not look perfect.

patrickrobrecht:
Fixed in #71

PR #67, opened 2018-03-04
MIgration from Raphaël JS to Chartist.JS

stklcode:
This PR is a partially rewritten JavaScript of the dashboard widget. As the title implies the graphics library has been replaced by [Chartist.JS](http://gionkunz.github.io/chartist-js/).

The new libaray 
* is slightly smaller (10kB GZIP excl. plugins),
* creates also SVG graphics (no loss of scalability),
* requires cleaner custom JS code, as the chart structure is entirely generated by the lib and
* separeted JS funcitonality from CSS styling.

Current state is more or less a drop-in replacement which reflects same functionality and about the same esthetics. Nevertheless it introduces new opportunities for tweaks and extensions.

The PR adresses #1 (German: replace Raphael with smaller lib) and also partially #66 (Chart Legibility), as chart scaling, clipping and zooming might be possible solutions for larger datasets (which would have been quite a pain with the old code)


bueltge (approved):

bueltge:
Great request! :+1:

bueltge:
We should use this change for switch to 1.6.0 of the plugin version. Maybe you @stklcode  can prepare the new version, changes in the root file, the readme or more?

patrickrobrecht:
I'll merge both open PRs later and update CHANGELOG and version number to 1.6.0 as proposed in our Slack channel.

PR #70, opened 2018-03-06
Drop minimum requirements from FAQ

chesio:
Both required PHP version and required WP version are prominently displayed in sidebar on plugin page. The information in FAQ is redundant and prone to become outdated (as it is now...)


stklcode (approved):
+1 for that, safe to remove

bueltge (approved):

patrickrobrecht:
The WP version in the FAQ was even false...

PR #75, opened 2018-03-15
Bundle JS source maps for pre-minified assets

stklcode:
Fixes issue report in the WP support forums: https://wordpress.org/support/topic/error-failed-to-load-resource-2/

The pre-minified Javascript fiels for Chartist and the tooltip plugin reference source maps
```
//# sourceMappingURL=chartist.min.js.map
```
which leads to 404 errors when accessing the admin panel with dev-tools active, because the `.map`files are not shipped.

This "bug" does not affect normal users, only admins with debug tools, so we could probably remove the maps from the distributable files, but for the same reason they wont't hurt (usually not requested), so i think it's fine to just bundle the extra ~250kB and not modify third party assets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea/
 css/*.min.css
 js/*.min.js
+js/*.min.js.map
 vendor/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file. This projec
 
 ## unreleased
 * Scaled datapoint size to number of records in dashboard widget to improve legibility 
+* Added JS source maps to avoid warnings with developer tools 
 
 ## 1.6.0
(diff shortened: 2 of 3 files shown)


patrickrobrecht:
Isn't there any solution removing the `//# sourceMappingURL=chartist.min.js.map` line from the minified files? That would be the cleaner solution in my opinion.

stklcode:
Sure, more than one
* pipe the JS through a filter
* stripping the last line (potentially dangerous when updated files have different layouts) 
* minify the minified file again, preserving only the top comment
* minify the original script without emitting maps

Drawback of changing is that we have to manipulate third party assets and potentially keep track of changes. E.g. they might use another minifier which generates `/* @source... */` syntax or places the comment inline in a 1-line JS file..
Last way of minifying the original is probably the safest way, but leads to other potential problems, as some GH issues for the tooltiip plugin state incompatibilities with some tools.

bueltge (approved):

PR #85, opened 2018-05-19
Rework Code-Style

stklcode:
As the title indicates, this PR adds some changes to the code-style.

First commit reworks comments to match WP conventions, second add some missing `esc_url()` and `esc_attr()` calls.

In total 4 PHPCS rules are re-enabled.

**For discussion:**
My fork goes one step further and renames all class-files to match the standard (https://github.com/stklcode/statify/commit/5cba94d82963fc2143f1f5184962e3a656424f60). Too much for now or just go for it?


patrickrobrecht:
If one starts to adjust the code style, we should do it consequently, i. e. fix as much as possible.

stklcode:
You asked for it, here it is... Renamed class files and removed all uses of `@codingStandardsIgnore` by sanitizing the superglobals.

patrickrobrecht (approved):
Good work!

PR #95, opened 2018-05-25
Refactor filter_input to filter_var for SERVER variables

stklcode:
`filter_input()` seemingly returns `NULL` on `INPUT_SERVER` values for some PHP implementations. This is caused by the way, superglobal variables are injected, esp. in CGI instanes. There are even PHP 7.x versions out there affected by this issue. (Bug report, e.g. https://bugs.php.net/bug.php?id=49184)

Solution is to partially reverted a change from #85 and use `filter_var()` on the `$_SERVER['...']` value instead. Doesn't look really handy, but whitouht raising a notice for non-existing value and a the null coalescing operator unavailable in 5.x is provides about the same type safety.

Issue and hint originally reported here (German): https://wordpress.org/support/topic/bei-mir-auch-seit-update-keine-zahlung-mehr/

### DIFF
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -37,14 +37,20 @@ class Statify_Frontend extends Statify {
 			$target   = urldecode( get_query_var( 'statify_target' ) );
 			$referrer = urldecode( get_query_var( 'statify_referrer' ) );
 		} elseif ( ! $use_snippet ) {
-			$target = filter_input( INPUT_SERVER, 'REQUEST_URI', FILTER_SANITIZE_URL );
+			$target   = filter_var(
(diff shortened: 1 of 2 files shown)


websupporter (approved):
looks good to me. cloned the branch and checked it a bit locally. the visits get tracked, although I couldn't reproduce the original problem leading to these changes, on a normal install, it works fine using `$_SERVER`

patrickrobrecht:
The same for me: Everything works fine as it did before. As I also couldn't reproduce the original problem, I asked I asked for some feedback in the support forums. I suggest to wait until someone whose environment is affected by the issue approves the changes. Then we'll release a bugfix version 1.6.3.

bueltge (approved):
We should use different commits for code and rules, phpcs in this context for the future to see the difference, history in the commits.

PR #98, opened 2018-06-01
Restore PHP 5.2 compatibility

stklcode:
In v1.5.0 we raised the minimum PHP requirement to 5.3.
In v1.5.3 (e1280a9 to be precise) we again raised the minimum PHP version to 5.6 - apparently for no reason whatsoever.

The code is still compatible with 5.3 today and with a single change, replacing a single use of `__DIR__` in the plugin's main file, 5.2 compatibility can be restored.

Of course I don't wanna encourage anybody to use historical software, but as recent tests have shown there are at least 5.4 and 5.5 users out there using Statify and we don't necessarily have to confuse them with pointless compatibility declarations....

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "statistics"
   ],
   "require": {
-    "php": "^5.6|^7",
+    "php": "^5.3|^7",
     "npm-asset/chartist": "^0.11.0",
     "npm-asset/chartist-plugin-tooltips": "^0.0.18"
   },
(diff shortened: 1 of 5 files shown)


patrickrobrecht:
The point when raising the declared required PHP version was that nobody of us tests Statify with versions < 5.6 such that we would not notice if something does not work with PHP 5.5 or lower (except if it can be detected by some automatic check).

If compatibility with PHP 5.2 is restored, the PHP version setting in the `readme.txt` could be set to `Requires PHP: 5.2` instead of removing it (to inform users about the compatibility).

bueltge (approved):

chesio (approved):

PR #71, opened 2018-03-07
Dynamic radius of datapoints in dashboard widget

stklcode:
This PR adresses #66.
The radius of the data points (currently 4px) is now dynmically adjusted according to the actual number of datapoints.

| # of points | radius |
| ----------- | ------ |
| >365        | 0px    |
| >180        | 1px    |
| >90         | 2px    |
| <=90        | 4px    |

This results in a dashboard like this:
dynamic_pointsize

PR #101, opened 2018-09-12
scrollable chart

manumeter:
trying to make stats more readable even when users keep data for more than a couple of weeks by making the chart horizontally scrollable instead of squeezing it.

before:
screenshot from 2018-09-12 10-59-58

after:
screenshot from 2018-09-12 11-00-47
screenshot from 2018-09-12 11-00-54

it would be nice if you'd consider to accept at least the scroll-part of this pull request.


stkjj:
like the idea!

stklcode:
:+1:  for the idea, but the current implementation still requires some work. For smaller frames (e.g. 60 days shown below - default 14 would be even worse) this does not look nice to me:

pr101_smallchart

* the chart should always fill the widget
* scrollbars should not be shown if not needed, i.e. defeine an "acceptable" amount of datapoints (60, 90, ...?)
* I'd personally not add 7 ticks, maybe 3 (25%, 50%, 100%), but that should not be a blocker

manumeter:
good points. I've tried to achieve this with my next [commit](https://github.com/pluginkollektiv/statify/pull/101/commits/eedb81929936126ad4d851741d426850b162cbaa).

for less than 90 datapoints:
screenshot from 2018-09-12 13-04-54

for more than 90 datapoints:
screenshot from 2018-09-12 13-05-15

stklcode:
LGTM, thanks for your work.

However I would wait with a merge decision until #111 and #72 are finally discussed to get a complete package of reasonable changes.

manumeter:
Any news on this? The charts of my WP instances are growing more unreadable daily.

bueltge (approved):
Like it, will wait for the review from @stkjj

js/dashboard.js
```diff
@@ -33,20 +33,27 @@
 	}, {
 		low: 0,
 		showArea: true,
-		fullWidth: true,
+		fullWidth: ( data.length < 90 ),
+		width: ( data.length < 90 ? null : 5 * data.length ),
```
stklcode: I've had a look at the changes several times yet while updating to current master and had my cursor on the _Approve_ button already...

Now I've found one border case where we waste some space:

statify_101_wastedspace

Anything against a dynamic limit instead of "90 points"?

```javascript
	if ( jQuery( '#statify_chart' ).width() < data.length * 5 ) {
		fullWidth = false;
	}

	/* ... */

		fullWidth: fullWidth,
		width: ( fullWidth ? null : 5 * data.length ),
```

?

js/dashboard.js
```diff
@@ -56,14 +63,7 @@
...
 
-	pointRadius = 4;
-	if ( data.length > 365 ) {
-		pointRadius = 0;
-	} else if ( data.length > 180 ) {
-		pointRadius = 1;
-	} else if ( data.length > 90 ) {
-		pointRadius = 2;
-	}
+	pointRadius = 2;
```
patrickrobrecht: For displaying only a small dataset (e. g. the last 14 days) only I like the display with `pointRadius = 4`. I think this code should only be changed for the `data.length > 365` case (due to the tooltip not being shown if there's no point to hover)

stklcode:
Added another last-minute change.

* compare against real container size instead of using 90 points (as suggested by myself)
* use 4px datapoints, if they fit the chart (as requested by @patrickrobrecht)
* fall back to 2px point radius if 4px does not fit
* make chart scrollable, if 2px still does not fit

So more or less a combined approach of my dynamic point radius (#71) and scrollable chart (this PR).

(I know `#points * 8` is not a perfect match for "4px does not overlap", but should be close enough)

patrickrobrecht (approved):

Issue #100, opened 2018-07-31
Long loading time for our JS snippet


Zodiac1978:
>1s loading time for or XMLHTTPRequest `/?statify_referrer=&statify_target=%2F `
https://wordpress.org/support/topic/ladezeit-problem/

This is an asynchronous request, so this shouln't slow down the website, but why is this happening at all?

![bildschirmfoto 2018-07-31 um 13 50 14](https://user-images.githubusercontent.com/870475/43457991-cc40bfa0-94c8-11e8-80c8-2db7e9c0e9d1.png)

![bildschirmfoto 2018-07-31 um 13 50 36](https://user-images.githubusercontent.com/870475/43457980-bd3f2c58-94c8-11e8-812e-8dfda6cdf5cf.png)

stklcode:
Can't reproduce this issue on multiple sites, RTT always in range of 50-100ms...

Best guess is that some other plugin is doing heavyweight operations before Statify kicks in in the `plugins_loaded` phase. Page generation time without a cache (seems _WP Fastest Cache_ is active) would be intersting here.

As also mehtioned in the forums, the targeted site has an average page load time (TTFB) of over 1s (tops out around 2.5s) for yet uncached pages (observed by clicking through the menus). I'd say slow server or heavy plugin activities and not a Statify issue.

stklcode:
Based on my observations, I don't think we have a major performance issue here.

Nevertheless, we could investigate in refactoring or JS tracking to e.g. the AJAX endpoint to not waste time for other plugins and operations when tracking is the only action of interest.

Zodiac1978:
@stklcode Should this ticket stay open for the refactoring idea or do you want to open another ticket for this?

stklcode:
Hard to tell. Actually the reported "behavior" is not really an error. But what else would we write into a new issue, so I'd keep it open for now...

I played around with switching to AJAX API calls couple of weeks ago that reduces load time of the JS call. Obvious, most plugins opt-out or at least do much less work on these calls and the template is not parsed. I'll prepare a PR for discussion on that soluition.

patrickrobrecht:
After merging #109 a WP AJAX call is used, which hopefully fixes the problem.

PR #109, opened 2018-10-02
Refactor JS snippet to use WP AJAX

stklcode:
The current JavaScript tracking method issues a GET-request to the home page with two parameters, `statify_referrer` and `statify_target`. As a result the usual WP page is processed, due to the parameters without caching (at least with the most comon caching plugins).

As mehtioned in #100 the tracking call might take long time depending on the actual setup (seen times up to 2s). That's because the request is intercepted and completed after Statify has been intialized. And that is typically not the first plugin loaded.

This PR replaces the custom GET tracking method by a POST call to a WP AJAX action. On AJAX calls the overhead is far less than on actual page generation, because many plugins opt-out immediately.

PS: I already tagged some changes in the PHP code with `@since 1.7.0`, because I don't see this change to appear in any 1.6.x release obviously. Doesn't mean anything however, might be rescheduled to any milestone :smiley:


statify.php
```diff
@@ -20,6 +20,7 @@
 define( 'STATIFY_FILE', __FILE__ );
 define( 'STATIFY_DIR', dirname( __FILE__ ) );
 define( 'STATIFY_BASE', plugin_basename( __FILE__ ) );
+define( 'STATIFY_VERSION', '1.6.3' );
```
Zodiac1978: Shouldn't this be 1.7.0?

stklcode: Absolutely (if we don‘t add this to 1.6.x - and I don‘t think so - or postpone this to 1.8 obviously)

js/snippet.js
```diff
@@ -1,11 +1,13 @@
...
-	const statifyReq = new XMLHttpRequest();
-	statifyReq.open(
-		'GET',
-		document.getElementById('statify-js-snippet').getAttribute('data-home-url')
-		+ '?statify_referrer=' + encodeURIComponent(document.referrer)
-		+ '&statify_target=' + encodeURIComponent(location.pathname + location.search)
-	);
-	statifyReq.send(null);
-} catch (e) {
+	jQuery.ajax( {
```
patrickrobrecht: Can we rely on jQuery being loaded on every WP site? Just because WP loads jQuey by default, this does not mean that a theme may change this behavior, right?

stklcode: No, we shouldn't. Sure around 99% (guess) of all WP sites have jQuery active, but I for myself disabled it for non-admin users on multiple sites, so I wonder why I've used it here at all...

Reverted back to vanilla JS in the next commit.

inc/class-statify-frontend.php
```diff
@@ -22,42 +22,42 @@ class Statify_Frontend extends Statify {
...
+			$referrer = isset( $_SERVER['HTTP_REFERER'] ) ? wp_unslash( $_SERVER['HTTP_REFERER'] ) : '';
 		} else {
 			return false;
 		}
 
+		// Sanitize.
+		$target = filter_var( $target, FILTER_SANITIZE_URL );
+		if ( is_null( $target ) || false === $target ) {
+			$target = '/';
+		} else {
```
bueltge: The `else` is not really necessary, so you should delete that and use only the var if not.

inc/class-statify-frontend.php
```diff
@@ -116,6 +116,20 @@ public static function track_visit() {
 		return self::_jump_out( $is_snippet );
 	}
 
+	/**
+	 * Track the page view via AJAX.
+	 *
+	 * @return void
```
bueltge, line 127: In my mind there is not helpful. Also the hint on phpdoc - This tag should not be used for constructors or methods defined with a void return type.

bueltge (changes requested):

bueltge:
Hi @stklcode 
great PR, thanks for your time to work again on the plugin. I have only two small hints, not important, but maybe helpful for you to merge this PR.

inc/class-statify-frontend.php
stklcode: We're using this tag in 15 places... Apparently all introduced by myself (default pattern in my IDE because it's mandatory in most of my styleguides - and btw. the PHPdoc ref marks this as "MAY be omitted").

We shouldn't mix it up, so either remove all or none.
Along with some int(eger) and bool(ean) tags probably little out of scope of this PR.

inc/class-statify-frontend.php
stklcode: Good point, not it's gone.

inc/class-statify-frontend.php
bueltge: Yes, plugincollektiv need a style guide, simplest way is the WP codex. In this context it is only a hint and the result of different views, works outside of the pluginkollektiv projects. Feel free to leave that or remove all ;)

bueltge (approved):

Issue #106, opened 2018-10-02
Add option on settings page to show stats for different roles


Zodiac1978:
A filter is sometimes scary for beginners. Why not adding this as an option on the settings page?
https://github.com/pluginkollektiv/statify/wiki/en-Hooks#statify__user_can_see_stats

Are there any reasons why this should not be done?

stklcode:
Can't see any point against that. Would not overload the dashboard Widget however, yet another point for adding a dedicated settings page.

Zodiac1978:
> yet another point for adding a dedicated settings page.

People sometimes miss the configuration page at all (see: https://www.internetkurse-koeln.de/wordpress-statistik-statify-versus-google-analytics-und-co/ at the bottom "Nachtrag").

How about moving the whole setting for the dashboard widget/Statify to an own settings page?

As a fallback we could leave a link to the new settings page on the dashboard widget config page.

stklcode:
> How about moving the whole setting for the dashboard widget/Statify to an own settings page?

Like so?

![statify_106_screenshot2](https://user-images.githubusercontent.com/12963621/46407169-3bb53100-c70e-11e8-83cc-a526e4be20b2.png)
![statify_106_screenshot1](https://user-images.githubusercontent.com/12963621/46407168-39eb6d80-c70e-11e8-8de8-256cf23207c2.png)

(not a mockup, fully funcitonal local fork)

Zodiac1978:
Wow. Just wow! 💯 🥇

Zodiac1978:
What about using this UI for our settings:

<img width="651" alt="Bildschirmfoto 2021-11-13 um 20 52 50" src="https://user-images.githubusercontent.com/870475/141657298-a63e546c-f8b6-499d-b7d6-d576c4e05de6.png">

The list of user roles is, of course, dynamic and lists all registered user roles in the current WP install.

RalfRoemling:
Anything new on this one? I have exactly this usecase not working at the moment and would have to escalate roles of my Editors to Admin.

Zodiac1978:
@RalfRoemling If you are comfortable to add some code (via functions.php, functionality plugin or code snippets plugin, you can use the code mentioned on the hook documentation:
https://statify.pluginkollektiv.org/documentation/hooks/#user_can_see_stats

RalfRoemling:
Yep, read that already in the older posts but a GUI solution would be more comfy and clearer in it's intention :-)

[...]

Issue #103, opened 2018-09-28
Re-Think skip tracking filter


Zodiac1978:
I'm not convinced of the usage of the skip tracking filter.

For example: If someone wants to track a Membership site (or community site with BuddyPress) it would be necessary to track logged in visitors. But our approach is to exit the skip_tracking function if the filter is not null.

The user needs to set the filter *and* he/she needs to copy all the other code lines to just remove `is_user_logged_in()` which is not as easy as it could be. 

https://github.com/pluginkollektiv/statify/blob/master/inc/class-statify-frontend.php#L139-L176

For example:
```
add_filter(
    'statify__skip_tracking',
    function() {
        if ( is_logged_in() ) {
            // Track logged in users, excluding feed, preview, error and search (skip otherwise)
            return !is_feed() && !is_preview() && !is_404() && !is_search();
        }
	
        // Continue regular processing.
        return null;
    }
);
```

Zodiac1978:
More context: https://wordpress.org/support/topic/logged-in-users-12/

2ndkauboy:
I agree that the function should be rewritten. There should only be one condition or conditional function in each if block and all of then should be able to be disabled using a filter. And additional filter at the end should allow skipping the teaching due to other conditions.

stklcode:
Agree on that, however a user should be able to white and blacklist on arbitrary conditions, that's what the hook is for in my opinion.

To not overload the existing hook, we might consider adding a second hook to filter the default settings. like `statify__filter_opts( array $opts )`.

Thinking further, an array of 9+ booleans is not what comes to my mind when I think of comfort. What about adding a settings page checkboxes instead?
This would really enable the average user to modify the behavior, brings transparency and we can add short info texts to explain each condition.

Zodiac1978:
> What about adding a settings page checkboxes instead?

I think that's a great idea!

I stumbled upon another related issue in the support forums (about the first block "Skip tracking via User Agent."):
> Statify comes with a default filter that whitelists Windows, Linux, iPhone, etc.
> The mobile bots however match this to force mobile display instead of desktop version.
>
> Mozilla/5.0 (iPhone; CPU iPhone OS 8_1 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12B411 Safari/600.1.4 (compatible; YandexBot/3.0; +http://yandex.com/bots) — Indexing robot.
>
[...]

PR #111, opened 2018-10-03
Introduce settings page and more options

stklcode:
This PR covers multiple issues and improvements. Feel free to edit or add.

Most significant change is the introduction of a new settings page. The widget backview is reduced to the two widget-related options (about top lists). I added a link to jump directly to the new page to not annoy existing users that are used to the old config panel.

Second I introduced three new checkboxes to control the `skip_tracking` behavior. I added the existing blacklist option to this group alongside with "logged in users", "feed" and "search". This partially covers #103, however the remaining options are not controllable yet. Might be combined with @2ndkauboy's solution to allow total control for pro users.
Especially the user agent filter with the RegEx might be suitable to configure here, but this is more of and advanced feature and not that easy to understand.

With the new page, the role filter (#106) can be integrated smoothly.
PR #72 is somewhat obsoleted or has to be adapted to the new method.

Finally two screenshots of the results:
statify_106_screenshot2
statify_106_screenshot1

### DIFF
--- a/inc/class-statify-settings.php
+++ b/inc/class-statify-settings.php
@@ -136,7 +136,7 @@ class Statify_Settings {
 	public static function options_snippet() {
 		?>
 		<input id="statify-snippet" type="checkbox" name="statify[snippet]" value="1" <?php checked( Statify::$_options['snippet'], 1 ); ?>>
-		(<?php esc_html_e( 'Default', 'statify' ); ?>: <?php esc_html_e( 'No' ); ?>)
+		(<?php esc_html_e( 'Default', 'statify' ); ?>: <?php esc_html_e( 'No', 'statify' ); ?>)
 		<br>
 		<p class="description"><?php esc_html_e( 'This option is strongly recommended if caching is in use.', 'statify' ); ?></p>
 		<?php
@@ -175,7 +175,7 @@ class Statify_Settings {
 	public static function options_today() {
 		?>
 		<input  id="statify-today" type="checkbox" name="statify[today]" value="1" <?php checked( Statify::$_options['today'], 1 ); ?>>
-		(<?php esc_html_e( 'Default', 'statify' ); ?>: <?php esc_html_e( 'No' ); ?>)
+		(<?php esc_html_e( 'Default', 'statify' ); ?>: <?php esc_html_e( 'No', 'statify' ); ?>)
 		<?php
 	}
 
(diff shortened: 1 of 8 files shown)


patrickrobrecht (commented):
Looks quite good. I've just added a commit fixing the text domain and changing some of the descriptions. I'm wondering whether the settings title can be put into `<label>` for the corresponding `input`s - `add_settings_field` doesn't seem to have an option for that. WP seems to do that on the WP settings pages (except for checkboxes).

Maybe the "skip tracking" options could be displayed as the options on the WP "Discussion" settings page.

stklcode:
> I'm wondering whether the settings title can be put into `<label>` for the corresponding `input`s - `add_settings_field` doesn't seem to have an option for that.

Sure it does: https://developer.wordpress.org/reference/functions/add_settings_field/

> $args
>(array) (Optional) Extra arguments used when outputting the field.
>
>* 'label_for'
> (string) When supplied, the setting title will be wrapped in a <label> element, its for attribute populated with this value.
>* 'class'
> (string) CSS Class to be added to the <tr> element when the field is output.
>
>Default value: array()

Didn't know that until 5 minutes ago, but it obviously works, see latest commit :smiley:

patrickrobrecht (approved):

inc/class-statify-settings.php
```diff
@@ -0,0 +1,258 @@
...
+	/**
+	 * Option for tracking via JS.
+	 *
+	 * @return void
+	 */
+	public static function options_snippet() {
+		?>
+		<input type="checkbox" name="statify[snippet]" id="statify_snippet" value="1" <?php checked( Statify::$_options['snippet'], 1 ); ?>
+			title="<?php esc_attr_e( 'Tracking via JavaScript', 'statify' ); ?>">
+		(<?php esc_html_e( 'Default', 'statify' ); ?>: <?php esc_html_e( 'No' ); ?>)
```
Zodiac1978: We should never use core translations. If they change we lose this translation. For consistency and completeness we should always use our own text domain.

stklcode: Okay, let's keep consistent :+1:

stklcode:
The admin menu is yet limited to the `manage_options` privilega, however the links are shown with `edit_dashboard`. Bad UX when clicking a link ends up in a _permission denied_ page.

https://github.com/pluginkollektiv/statify/blob/a2b533009b9884d77051e53aca22d55ee3bebd22/inc/class-statify-settings.php#L289-L295

I separated the 3 possible permissions:
* See dashboard widget - `edit_dashboard` or filter `statify__user_can_see_stats`
* Edit dashboard widget - `edit_dashboard`
* Edit all settings - `manage_options`

PR #116, opened 2018-10-24
Issue #110 "JS tracking and AMP compatibility"

tthemann:
First draft of an out-of-the-box support for the Automattic AMP plugin (see https://github.com/pluginkollektiv/statify/issues/110). Adding analytics using the amp_post_template_analytics (see https://github.com/Automattic/amp-wp/wiki/Analytics) seems to be the only easy and documented way to implement amp-analytics programmatically. Might need some cleanup (as I´m not really familiar with the WordPress coding standards) and some more testing in different environments.


inc/class-statify-frontend.php
```diff
@@ -318,4 +318,34 @@ public static function wp_footer() {
...
+		if ( ! is_array( $analytics ) ) {
+			$analytics = array();
+		}
+
+		$analytics['statify'] = array(
+			'type' => 'statify',
+			'attributes' => array(),
+			'config_data' => array(
+				'requests' => array(
+					'event' => get_site_url() . '/?statify_referrer=${documentReferrer}&statify_target=${canonicalPath}amp/'
```
bueltge: Maybe we should sanitize this url, like `esc_url()`.

bueltge (changes requested):
Looks great, coding standard also. However I'm not familiar with AMP and the AMP plugin.

stklcode:
If JS tracking is disabled, at least the first AMP call triggers two tracking calls. I pushed a fix to not add any AMP analytics spec if non-JS tracking mode is active.
Most likely a really rare case, but double tracking is not what we want here.

----

I also removed the `type` attribute. _Statify_ is no registered analytics vendor and we don't intend to become one I believe, so the AMP analytics script always receives a 404 trying to trigger `https://cdn.ampproject.org/rtv/.../v0/analytics-vendors/statify.json`

inc/class-statify-frontend.php
stklcode: We probably shouldn't. The placeholders `${documentReferrer}` and `${canonicalPath}` are AMP builtins and replaced on client side and all other parts are either WP builtins or hardcoded.

Generated JSON by the AMP plugin:
```json
{
  "requests": {
    "event": "https:\/\/example.com\/?statify_referrer=${documentReferrer}&statify_target=${canonicalPath}amp\/"},
    "triggers": { "trackPageview": { "on": "visible", "request": "event", "vars": { "eventId": "pageview" } } 
  }
}
```

Escaping the braces here would yield something like `.../?statify_referrer=$documentReferrer&#038;statify_target=$canonicalPathamp/` which no longer provides the desired result.

stklcode (approved):
LGTM.
Solution does not interfere with non-AMP sites and works as expected for my local AMP testbed.

I've added a few changes:
* some minor code style issues
* correction for double tracking if JS is disabled
* removed the AMP analytics vendor type

stklcode:
Note: re-test in combination with WP-AJAX refactoring #109, as _Statify_ itself issues a POST request to AJAX endpoint instead of the GET call used for AMP. Might need to append `.transport.xhrpost = true` flag to the configuration object and add nonce to vars.

stklcode:
Refactored the analytics configuration to new AJAX hook for tracking.

I'm not 100% sure if `${canonicalPath}amp/` is correct in any case. However it works for me.
We counld even discuss whether to drop the `/amp/` suffix completely, so AMP and classic calls are treated equally.

I'm fine with the current solution for now. However too much involved, so somebody else should have a look at the implementation.

inc/class-statify.php
```diff
@@ -89,6 +89,9 @@ public function __construct() {
 			add_action( 'template_redirect', array( 'Statify_Frontend', 'track_visit' ) );
 			add_filter( 'query_vars', array( 'Statify_Frontend', 'query_vars' ) );
 			add_action( 'wp_footer', array( 'Statify_Frontend', 'wp_footer' ) );
+			if ( function_exists( 'is_amp_endpoint' ) ) { // Automattic AMP plugin present.
```
bueltge, line 98: I would see this statement, logic in a separate method and not logic inside the constructor. For me is more clear and on focus to switch to modern development we need to go to an empty constructor in context of usage from unit tests. You agree @tthemann and would you enhance them? So only fire this method currently via the constructor and the logic in a enhanced method.

bueltge (changes requested):

inc/class-statify.php
stklcode: The constructor already does heck of a lot of stuff a constructor should not do...

I think we should eliminate the pointless `_constructor()` / `static instance()` combination entirely in favor of a single static `init()` method. _But not in the context of this PR._

**Edit:** addressed this point in #140, considering this discussion resolved.

tthemann: To be honest, these lines have been almost the only PHP code I wrote since the early 2000s - therefore I might not be the right person to discuss coding standards or philosophy ;-).

bueltge (approved):

inc/class-statify.php
bueltge: PR #140 solve this so we should merge this after update for the new code base after merge #140.

bueltge (approved):

PR #117, opened 2018-11-05
Fix target and referer reporting when PHP and database differ on timezone

sophiehuiberts:
My hosting provider has strftime in UTC and DATE(NOW()) in Amsterdam time.
This broke top referer and top target reporting when reporting today's
top, for 1 or 2 hours per day, depending on daylight savings time.


stklcode (approved):
I resolved the merge conflicts with #134 (which also requires "today") and slightly reworked the DB queries to use prepared statements instead of string concatenation.

Now we're using PHP server consistently. LGTM.

bueltge (approved):

PR #134, opened 2019-10-01
Add option to show total visits

### LINKED ISSUE #128: Show number of total views tracked by Statify
A client asked if it is possible to see the total number of views for a website (so the total count of views that is in the Statify table).

Is that something worth adding, maybe hidden behind an option?

yurihs:
Hi! This should resolve #128.

Please let me know if there are any problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. This projec
 ## Unreleased
 * Introduced separate settinge page and reduced widget backview to widget settings only
 * Add options to track logged in users, feeds and search requests
+* Add option to show total visits
 
 ## 1.6.3
 * Fix compatibility issue with some PHP implementations not populating `INPUT_SERVER`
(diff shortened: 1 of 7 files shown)


stklcode:
Thanks very much @yurihs for this work.
I've only had a look into the code so far. Implementation looks good to me.

One additional suggestion though:
"all time" is quite a flexible time range. Might be the last 5 days or the last 5 years depending on configuration and actual traffic.
What about adding the information "since [date]" , either instead or in addition of "all time"?

yurihs:
That makes sense! Please have another look @stklcode.

stklcode:
Great! I've added a minor CSS correction for alignment with the pre-existing tables.

Works fine on top of the current master. With #72 merged, `$stats['visits'][0]['date']` is no longer correct, because this would display the overall totals from the database along with the oldest date from the display range.

Feel free to work around this, but I have to rework my PR anyway, so I can take care of that there.

stklcode (approved):

PR #135, opened 2019-11-11
Add GitHub Actions for automated deployments

2ndkauboy:
To be merged before the next release.


.gitattributes
```diff
@@ -1,10 +1,15 @@
-* text eol=lf
-*.jpg -text
-*.png -text
-*.gif -text
```
bueltge: I mean we should leave them because we get several problems with media files inside the repository if we remove them.

bueltge (commented):
@2ndkauboy small question or note, maybe you can validate them and after that we should merge them. @krafit

.gitattributes
```diff
@@ -1,10 +1,15 @@
...
-.stylelintrc.json            export-ignore
+# Files
+/.eslintrc.json export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.stylelintrc.json export-ignore
+/.travis.yml export-ignore
+/composer.json export-ignore
+/composer.lock export-ignore
+/phpcs.xml export-ignore
```
patrickrobrecht, line 19: I think that the files `CHANGELOG.md`, `package.json`, `package-lock.json` and `README.md` (we have a separate `readme.txt` for wp.org) and the directories `node_modules` and `vendor` should be added as `export-ignore`, right?

patrickrobrecht (changes requested):

.gitattributes
stklcode: `node_modules` and `vendor` are already listed above under _directories_

Changelog is at least not completely useless, but not really necessary either.

patrickrobrecht (approved):

.gitattributes
patrickrobrecht: Added the missing files to the `.gitattributes` file.

patrickrobrecht (approved):

PR #136, opened 2020-04-07
update devenv plus minor code style corrections

stklcode:
* update dev dependencies
* add PHP 7.3 and 7.4 to CI roster
* minor PHP code style corrections (indentation, parenthesis, ...)
* minor JS/CSS code style corrections to match WPCS
  * eslint/stylelint configuration files added to repo
     no automatic checks, didn't want to introduce NPM as second build tool and *lint tools are a pain to use with Composer


patrickrobrecht:
LGTM, but regarding the CSS / JS check: Without an automatic check this seems to be a little useless. I don't know a CSS / JS linter for Composer, but I don't see a problem with npm as a second build tool. We could integrate the commands into the `composer.json` such that npm is just running in the background.

I tried to include the dependencies via [asset-packagist.org](https://asset-packagist.org/), but that didn't work (got an error `Your requirements could not be resolved to an installable set of packages.` after adding `npm-asset/stylelint` to the dev dependencies.

stklcode:
Used to work out with older releases of the linters, but as said, current linter with asset-packagist and composer is a pain.

In TravisCI both tools are available by default (not necessarily on different platforms), so I embedded a block like this in other projects:
```json
    "lint-php": [
      "phpcs --standard=phpcs.xml -s"
    ],
    "lint-css": [
      "./node_modules/stylelint/bin/stylelint.js styles/stylesheet.css",
     ],
    "lint-js": [
      "./node_modules/eslint/bin/eslint.js scripts/script.js"
    ],
    "lint-all": [
      "@lint-php",
      "@lint-css",
      "@lint-js"
    ],
```
with hardcoded paths in _.*lintrc.json_ configs (from asset-packaginst solutions with older tool revisions). Executing with `npx` should do the job as well.

stklcode:
Integrated both linters with separate Composer scripts `lint-js` and `lint-css` and corresponding lines to  the CI descriptor to install NPM dependencies and run the checks.

I intentionally did not integrate the NPM tasks in Composer "post-install" hook to not break Composer-based WP setups (with Node not necessarily available) for code-style only reasons.

ESLint Plugin v4 requires npm 6.9 or later which is not available in default Travis PHP environments, so I declared dependency to v3 for now.

patrickrobrecht (approved):

PR #142, opened 2020-04-22
add separate AJAX action for tracking of logged in users

stklcode:
Fixes a regression from the AJAX refactoring combined with tracking of logged-in users.
Actions with "nopriv" prefix are not evaluated for those users and thus tracking is never actually executed.


inc/class-statify.php
```diff
@@ -79,6 +79,9 @@ public static function init() {
 
 		if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
 			add_action( 'wp_ajax_nopriv_statify_track', array( 'Statify_Frontend', 'track_visit_ajax' ) );
+			if ( 0 === self::$_options['skip']['logged_in'] ) {
+				add_action( 'wp_ajax_statify_track', array( 'Statify_Frontend', 'track_visit_ajax' ) );
+			}
```
patrickrobrecht, line 84: Adding line 83 is sufficient, isn't it? The condition is checked by the `track_visit_ajax` function.

To jump out as early as possible, one could move the skip tracking check to the very first line of track_visit, right?

stklcode: You are right.

`skip_tracking` however triggers external action hooks before evaluating our logic, even with moved the login check up.

Moving the skip-call up the `track_visit` method breaks  early return for the non-AJAX call which is likely triggered always for registered users.

patrickrobrecht (approved):

PR #144, opened 2020-05-01
skip tracking for favicon.ico redirects

stklcode:
Fixes WP 5.4 issue reported in WP support forums:
https://wordpress.org/support/topic/favicon-ico/

In WP 5.4 the favicon fallback has been refactored for more flexibility<sup>1</sup>.
Due to that behavior, plugins are now initialized before the redirect to the actual icon file and Statify tracks the visit.

Introduce another exclusion, s.t. those calls are treated as "internal" and thus skipped. Utilizes the newly introduces `is_favicon()` check<sup>2</sup>.

[1] https://make.wordpress.org/core/2020/02/19/enhancements-to-favicon-handling-in-wordpress-5-4/
[2] https://developer.wordpress.org/reference/functions/is_favicon/

### DIFF
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -246,8 +246,9 @@ class Statify_Frontend extends Statify {
 	 * @return   boolean  $skip_hook  TRUE if NO tracking is desired
 	 */
 	private static function _is_internal() {
-		// Skip for preview, 404 calls, feed and search access.
-		return is_preview() || is_404() || is_feed() || is_search();
+		// Skip for preview, 404 calls, feed, search and favicon access.
+		return is_preview() || is_404() || is_feed() || is_search()
+			|| ( function_exists( 'is_favicon' ) && is_favicon() );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


inc/class-statify-frontend.php
```diff
@@ -247,7 +247,8 @@ private static function is_bot( $user_agent ) {
 	 */
 	private static function _is_internal() {
 		// Skip for preview, 404 calls, feed and search access.
```
Zodiac1978: This line could be expanded with the Favicon check info

Zodiac1978:
Looks good to me!

inc/class-statify-frontend.php
stklcode: Right. If we document self explaining code, we should do it completely :wink:

Zodiac1978: Or remove completely as this is a little bit silly ;)

Zodiac1978 (approved):
Thanks for fixing!

patrickrobrecht (approved):

PR #146, opened 2020-05-11
add .distignore for automatic deployment control

stklcode:
The deployment action utilizes `git archive` if no _.distignore_ is specified. Doing so, the minified JS/CSS resources are ignored, because not only _.gitattributes_ is processed, but due to _.gitignore_ the _*.min.[js|css]_ files have never been added to git and won't become part of the archive.

Using explicit _.distignore_ list we now separate git / Compsoer excludes from WP deployment. This triggers simple `rsync` without git logic.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,6 +8,7 @@
 /vendor export-ignore
 
 # Files
+/.distignore export-ignore
 /.eslintrc.json export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -8,6 +8,7 @@
 /vendor export-ignore
 
 # Files
+/.distignore
```
patrickrobrecht: Isn't the `export-ignore` missing here? The .distignore LGTM.

stklcode: Yes, you're right. Fixed with rebase.

patrickrobrecht (approved):

PR #148, opened 2020-05-11
Delete saved transient for the dashboard whenever the widget settings are changed via the settings page

### LINKED ISSUE #147: Transient data should be deleted when widget settings are changed via settings page
Hello,

kudos to all who worked on 1.7, great job! :)

I just noticed that when I change "Period of data display in Dashboard" via Settings page, it doesn't take effect until I resave it via widget configuration or wait until transient expires.

One way how to fix this is via [`update_option_{$option}`](https://developer.wordpress.org/reference/hooks/update_option_option/) hook:
```php
add_action( 'update_option_statify', function () { delete_transient('statify_data'); }, 10, 0);
```

patrickrobrecht:
Fixes #147

### DIFF
--- a/inc/class-statify-settings.php
+++ b/inc/class-statify-settings.php
@@ -24,7 +24,7 @@ class Statify_Settings {
 	 * @return void
 	 */
 	public static function register_settings() {
-		register_setting( 'statify', 'statify', array( __CLASS__, 'sanitize_options' ) );
+		register_setting( 'statify', 'statify', array( __CLASS__, 'save_options' ) );
 
 		// Global settings.
 		add_settings_section(
@@ -243,6 +243,21 @@ class Statify_Settings {
 		<br>
 		<p class="description"><?php esc_html_e( 'Enabling this option excludes any views of logged-in users from tracking.', 'statify' ); ?></p>
 		<?php
+	}
+
+	/**
+	 * Save the submitted options.
+	 *
+	 * @since 1.7.1
+	 *
+	 * @param array $options the submitted options.
+	 *
+	 * @return array Validated and sanitized options.
+	 */
+	public static function save_options( $options ) {
+		$validated = self::sanitize_options( $options );
+		delete_transient( 'statify_data' );
+		return $validated;
 	}
 
 	/**


chesio:
I personally prefer using the  [`update_option_{$option}`](https://developer.wordpress.org/reference/hooks/update_option_option/) hook. It has the benefit that it triggers on any call to `update_option('statify', ...)`, not only when settings page is processed, but from anywhere - including any 3rd party integrations or when [widget configuration is saved](https://github.com/pluginkollektiv/statify/blob/master/inc/class-statify-dashboard.php#L242). This way the subsequent explicit call to [delete transient](https://github.com/pluginkollektiv/statify/blob/master/inc/class-statify-dashboard.php#L245) could be removed.

stklcode:
The proposed changes look reasonable to me. 

However @chesio has a good point and we need to introduce another method anyway, if we don't want copy&paste here.


<details><summary>In <code>Statify::init()</code>, likely the <code>is_admin()</code> section.</summary>
<pre><code>
add_action( 'update_option_statify', array( 'Statify_Settings', 'delete_statify_transient' ) );
</code></pre>
</details>

<details><summary>In <code>Statify_Settings</code>, could also be In <code>Statify</code>...</summary>
<pre><code>
/**
 * Delete transient data to refresh the dashboard widget.
 *
 * @since 1.7.1
 *
 * @return void
 */
public static function delete_statify_transient( ) {
	delete_transient( 'statify_data' );
}
</code></pre>
</details>

----

**Pros:**
* cannot be forgotten
* compatible with other update ways (API, third-party plugin, ...)

**Cons:**
* Saving options wihtout changes cannot be (mis)used to update the widget anymore, because the action is only triggered, if anything has been changed.
(a "refresh" button in the widget seems more clear anyway)


----

**Just a side note:**
The [cachify_flush_cache](https://github.com/pluginkollektiv/statify/blob/master/inc/class-statify-dashboard.php#L248-L250) should also be moved out of `_save_widget_options()`, and into `save_options` because the widget options don't affect the cache anymore and changing JS tracking from the settings page should invalidate the cache.

patrickrobrecht:
Closed in favor of #154.

PR #154, opened 2020-05-13
add action after options change to delete transients and flush Cachify

### LINKED ISSUE #147: Transient data should be deleted when widget settings are changed via settings page

### LINKED ISSUE #149: Heavily more visitors since v1.7.0

### LINKED ISSUE #152: Cachify no longer flushed when changing JS settings
When settings are saved in the dashboard widget backview, we do flush the _Cachify_ cache, if both plugins are active.

https://github.com/pluginkollektiv/statify/blob/6d3910c780c4129aa57fe4fd57de94513822f1cc/inc/class-statify-dashboard.php#L247-L250

Since 1.7 the dashboard widget settings no longer affects any page markup and flushing the cache is useless in this case. But saving options on the settings page allows (de)activating Javascript tracking and the cache is not flushed in this case.

stklcode:
Supersedes #148.
Fixes #147 and fixes #152.

Introduce a new action hook for [`update_option_statify`](https://developer.wordpress.org/reference/hooks/update_option_option/) as suggested by @chesio.

In contrast to #148 this allows easy extension to get around the problem that Cachify cache was not flushed. This is now fixed with minor enhancement that it's only flushed if JS settings have actually changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file. This projec
 
 ## unreleased
 * Fix refresh of the dashboard widget when settings have been changed through the settings page (#147)
+* Fix _Cachify_ cache not being flushed after changing JavaScript settings (#152)
 
 ## 1.7.0
 * Fix JavaScript embedding when bots visit before caching (#84) (#86) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
 
+## unreleased
+* Fix refresh of the dashboard widget when settings have been changed through the settings page (#147)
+
 ## 1.7.0
 * Fix JavaScript embedding when bots visit before caching (#84) (#86) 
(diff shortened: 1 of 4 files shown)


chesio:
Looks good to me, thank you @stklcode!

patrickrobrecht (approved):

PR #155, opened 2020-05-13
Include punctuation in string

### LINKED ISSUE #153: Superflous period in description
> Das Aktivieren dieser Option schließt alle Aufrufe mit Verweisen aus der Kommentar-Blacklist vom Tracking aus..

We should be more consistent with periods. Including or not including them ...

https://github.com/pluginkollektiv/statify/blob/3c4cbce44b74b5569a46ccc4b991a01a2366327e/inc/class-statify-settings.php#L230

vs. 

https://github.com/pluginkollektiv/statify/blob/3c4cbce44b74b5569a46ccc4b991a01a2366327e/inc/class-statify-settings.php#L244

I don't have a strong opinion which is better, but we should choose one variant and stay with it to avoid double periods as seen now in the German translation.

Zodiac1978:
Fixes #153


stklcode (approved):
LGTM

According to the translation overview it seems to be the only affected sentence.

PR #164, opened 2020-06-17
Set stable tag to 1.7.2

patrickrobrecht:


stklcode:
If this PR was supposed to trigger the GH asset deployment action, I think the Composer step is required in the corresponding action, too. Without there is still a mismatch for the minified resources and third-party dependencies...

patrickrobrecht:
Yeah, I released the stable tag change manually. #165 should fix that for next time, right?

PR #165, opened 2020-06-17
Fix readme/asset update action 

patrickrobrecht:
This bugfix should avoid diff in missing minified files by running composer install - the postinstall action will generate the required files.


stklcode (approved):
Works for me.
Just ran the action locally in a pair of test containers (action + SVN copy) and some dummy change and it got deployed :+1: 

I've added a minor change to the FAQ section of the _readme.txt_ to get a proof it actually works on GH, too.

patrickrobrecht:
Thanks for testing! Then we can merge the PR after the [issues with plugin updates on wordpress.org](https://make.wordpress.org/plugins/2020/06/18/plugin-update-issues-18-june-2020/) are resolved.

PR #168, opened 2020-07-13
Add an option to deactivate the nonce check during JavaScript tracking

patrickrobrecht:
We got some reports that Statify tracks less views on WP sites which have a long caching duration enabled. The problem in those installations may be that the nonce (expires after ~24 h) is not valid anymore and the view is not tracked. We discussed that we could make the nonce check optional - this PR implements this idea.

### DIFF
--- a/inc/class-statify-settings.php
+++ b/inc/class-statify-settings.php
@@ -8,7 +8,7 @@
  * @since     1.7
  */
 
-// Quit ic accessed directly..
+// Quit if accessed directly..
 defined( 'ABSPATH' ) || exit;
 
 /**
@@ -48,6 +48,14 @@ class Statify_Settings {
 			'statify',
 			'statify-global',
 			array( 'label_for' => 'statify-snippet' )
+		);
+		add_settings_field(
+			'statify-snippet-nonce',
+			__( 'Require nonce for JavaScript tracking', 'statify' ),
+			array( __CLASS__, 'options_nonce' ),
+			'statify',
+			'statify-global',
+			array( 'label_for' => 'statify-nonce' )
 		);
 
 		// Dashboard widget settings.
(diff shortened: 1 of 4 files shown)


inc/class-statify-settings.php
```diff
@@ -142,6 +150,22 @@ public static function options_snippet() {
...
+	 * @since 1.7.3
+	 *
+	 * @return void
+	 */
+	public static function options_nonce() {
+		?>
+		<input id="statify-nonce" type="checkbox" name="statify[nonce]" value="1" <?php checked( Statify::$_options['nonce'], 1 ); ?>>
+		(<?php esc_html_e( 'Default', 'statify' ); ?>: <?php esc_html_e( 'Yes', 'statify' ); ?>)
+		<br>
+		<p class="description"><?php esc_html_e( 'Activate this setting if the caching time is longer than the nonce time.', 'statify' ); ?></p>
```
stklcode: Should we assume that the average administrator knows about AJAX behavior and nonce validity?

Maybe add an additional note on the observable misbehavior like _... longer than the nonce time and you experience “Forbidden“ errors and visits are not counted._

patrickrobrecht: I've updated the description. After the release I would add a detailed hint in our [docs](https://statify.pluginkollektiv.org/documentation/settings/#js-tracking).

bueltge (approved):
Die if Schachtelungen gefallen mir nicht, könnte man ggf. schöner lösen. Parallel ab und an return false, da würde ich in eine variable schreiben, damit möglichst immer der gleiche Type zurück gegeben wird. 
Aber im Sinne des Ganzen passt es, hoffe, es löst den Fehler.

stklcode (approved):
Works for me.

Changes look reasonable, backwards compatibility give, manual testing OK.

PR #173, opened 2020-07-25
change deployment branch from master to stable and update actions

stklcode:
Switch to branch "stable" as new target to deploy stable releases.

Update actions to their current versions which have changed from master to _stable_ or _v1_/_v2_, too.

This PR already targets the new default target branch _develop_.


patrickrobrecht (approved):
LGTM

PR #174, opened 2020-07-25
support "disallowed_keys" instead of "blacklist" for WP 5.5 compat

stklcode:
In _WordPress_ 5.5 the "comment blacklist" has been renamed to "disallowed comment keys" along with the corresponding option key<sup>[1](#fn1)</sup>.

We now support the new option, if available.
Also the frontend wording is adapted to match the new labels.
Private methods and comments changed where reasonable. Public methods and internal option names left untouched for backwards compatibility.

[<a name="fn1">1</a>] https://make.wordpress.org/core/2020/07/23/codebase-language-improvements-in-5-5/


patrickrobrecht (approved):
Great, then we can release Statify 1.8.0 with support for WordPress 5.5 already.

PR #175, opened 2020-07-25
Add a refresh button in the dashboard

### LINKED ISSUE #157: Refresh button for the dashboard widget
Up to 1.7.0, opening the options dialog and saving options w/o any change deleted the transients. During testing, I (mis-)used that as a "refresh statistics immediate" functionality. Since 1.7.1, the transients are only deleted if an actual change has been made and I´m missing the "refresh" (actually, I´m just too busy to wait 4 minutes ;-) ). Wouldn´t it be handy to have some sort of "refresh" link in the dashboard widget?

patrickrobrecht:
Fixes #157

With adding the refresh button I increased the caching time to 15 minutes.

### DIFF
--- a/inc/class-statify-dashboard.php
+++ b/inc/class-statify-dashboard.php
@@ -264,14 +264,18 @@ class Statify_Dashboard extends Statify {
 	 * @since   0.1.0
 	 * @version 1.4.0
 	 *
-	 * @return  array  $data  Data from cache or DB
+	 * @param bool $use_cache If true use data from cache, otherwise force retrieving data from the database.
+	 *
+	 * @return  array  $data  stats data from cache or database
 	 */
-	public static function get_stats() {
+	public static function get_stats( $use_cache = true ) {
 
-		// Get from cache.
-		$data = get_transient( 'statify_data' );
-		if ( $data ) {
-			return $data;
+		// Get from cache if enabled.
+		if ( $use_cache ) {
+			$data_from_cache = get_transient( 'statify_data' );
+			if ( $data_from_cache ) {
+				return $data_from_cache;
+			}
 		}
 
 		// Get from DB.
@@ -288,7 +292,7 @@ class Statify_Dashboard extends Statify {
 		set_transient(
 			'statify_data',
 			$data,
-			MINUTE_IN_SECONDS * 4
+			MINUTE_IN_SECONDS * 15
 		);
 
 		return $data;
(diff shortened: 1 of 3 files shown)


inc/class-statify-dashboard.php
```diff
@@ -264,14 +264,18 @@ private static function _get_version() {
 	 * @since   0.1.0
 	 * @version 1.4.0
 	 *
-	 * @return  array  $data  Data from cache or DB
+	 * @param bool $use_cache If true use data from cache, otherwise force retrieving data from the database.
+	 *
+	 * @return  array  $data  stats data from cache or database
 	 */
-	public static function get_stats() {
+	public static function get_stats( $use_cache = true ) {
```
stklcode: I would have inverted the parameter like `$refresh = false` here.

* the only call of this method is negated (`get_stats( ! $refresh )`)
* `$use_cache = false` indicates that the cache is not used at all, while it is always updated
* I prefer the semantics of forced refresh, because this guarantees a fresh result while indicating cache usage does not tell if the result is actually loaded from cache or if the cache has been empty

inc/class-statify-dashboard.php
```diff
@@ -264,14 +264,18 @@ private static function _get_version() {
...
+	public static function get_stats( $force_refresh = false ) {
 
-		// Get from cache.
-		$data = get_transient( 'statify_data' );
-		if ( $data ) {
-			return $data;
+		// Get from cache if enabled.
+		if ( ! $force_refresh ) {
+			$data_from_cache = get_transient( 'statify_data' );
+			if ( $data_from_cache ) {
```
stklcode, line 276: Just a note:
This always causes a refresh if no data is stored in the database. `get_transient()` returns `false` if no value is available while the value is an empty string, if we've stored `null` before.

The behavior has been the same before, so nothing is changed here.
We _could_ think about either using strict comparison (`false === $data_from_cache`) here ~~or - better imho - move the `set_transient()` call into the `(if ( ! empty( $data['visits'] ) ) { ... }` block and not store `null` at all.~~ (would break refresh functionality, if data is cached and has been entirely cleaned up in the meantime)

stklcode (approved):

inc/class-statify-dashboard.php
patrickrobrecht: In think causing a refresh if no data is stored in the database is intended: For example, visiting the dashboard the first time after installing Statify should trigger getting the latest data from the database (if there are some already). That's why I called the parameter `$force_refresh` and not only `$refresh` (as a `$force_refresh = false` still might lead to a refresh).

PR #176, opened 2020-08-01
Prepare the relase of version 1.8.0

patrickrobrecht:
Thanks to the adjustments made by #174 we can also declare compatibility with WordPress 5.5


stklcode (approved):
Works for me.

Tested with WP 5.4.2 and 5.5 RC1.

Left a minor change in the asset deployment action, but that's all I got here.

PR #182, opened 2020-08-28
fix AMP compatibility for Standard and Transitional mode (#181)

### LINKED ISSUE #181: Invalid markup by AMP plugin with enabled JavaScript tracking
Every time I save a post I get two validation errors in the [official AMP plugin](https://wordpress.org/plugins/amp/) with enabled JavaScript tracking:

> Invalid inline script
> Fehlercode
> DISALLOWED_TAG
> Ungültiges Markup
> <script … >
> Element-Attribute
> type	text/javascript
> id	statify-js-js-extra
> Element-Name
> script
> Übergeordnetes Element
> body
> Textinhalt
>  128 Bytes
> /* <![CDATA[ */
> var statify_ajax = {"url":"https:\/\/kittmedia.com\/wp-admin\/admin-ajax.php","nonce":"3d82495992"};
> /* ]]> */

> Invalid script snippet.min.js
> Fehlercode
> DISALLOWED_TAG
> Ungültiges Markup
> <script … >
> Element-Attribute
> type	text/javascript
> src	https://kittmedia.com/wp-content/plugins/statify/js/snippet.min.js
> id	statify-js-js
> Element-Name
> script
> Übergeordnetes Element
> body

Due to the fact that Statify is recommending using the JavaScript tracking method to track AMP pages as well, I would assume that it will work properly. However, as soon as this invalid markup gets removed (which is required in order to serve AMP pages), Statify isn’t able anymore to track any AMP page visits.

AMP version: 2.0.0 (already occurred in version 1.5.x)
Statify version: 1.8.0
WordPress version: 5.5.0

stklcode:
Standard and Transitional mode use the `amp_analytics_entries` hook, Reader mode uses `amp_post_template_analytics`.
We only added a filter for the latter, so AMP compatibility was only available in Reader mode.

### DIFF
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -410,7 +410,7 @@ class Statify_Frontend extends Statify {
 	/**
 	 * Add amp-analytics for Standard and Transitional mode.
 	 *
-	 * @see hAMPttps://amp-wp.org/documentation/playbooks/analytics/
+	 * @see https://amp-wp.org/documentation/playbooks/analytics/
 	 *
 	 * @param array $analytics_entries Analytics entries.
 	 */
@@ -475,6 +475,11 @@ class Statify_Frontend extends Statify {
 					'on'      => 'visible',
 					'request' => 'pageview',
 				),
+			),
+			'transport'      => array(
+				'beacon'  => true,
+				'xhrpost' => true,
+				'image'   => false,
 			),
 		);
 	}
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -404,7 +404,30 @@ class Statify_Frontend extends Statify {
 	}
 
 	/**
-	 * Add amp-analytics.
+	 * Add amp-analytics for Standard and Transitional mode.
+	 *
+	 * @see hAMPttps://amp-wp.org/documentation/playbooks/analytics/
+	 *
+	 * @param array $analytics_entries Analytics entries.
+	 */
+	public static function amp_analytics_entries( $analytics_entries ) {
+		if ( ! is_array( $analytics_entries ) ) {
+			$analytics_entries = array();
+		}
+
+		// Analytics script is only relevant, if "JS" tracking is enabled, to prevent double tracking.
+		if ( self::is_javascript_tracking_enabled() ) {
+			$analytics_entries['statify'] = array(
+				'type'   => '',
+				'config' => wp_json_encode( self::make_amp_config() ),
+			);
+		}
+
+		return $analytics_entries;
+	}
+
+	/**
+	 * Add AMP-analytics for Reader mode.
 	 *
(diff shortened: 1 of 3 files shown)


MatzeKitt:
This works for me. 👍🏻

stklcode:
Great. Thanks for the quick feedback.

I‘ll add a condition to remove the unnecessary scripts tomorrow (if nobody is faster), then we can rollout the fix.

stklcode:
I've added the necessary check within the `Statify_Frontend::wp_footer()` method. Along with this change we now check for `amp_is_request()` with fallback to 'is_amp_endpoint()` (deprecated in AMP plugin 2.0).

An earlier check, e.g. not register `wp_footer` hook at all in the `Statify::init()` phase is not possible this way, because the AMP check is not initialized that early, so we jump out in the very first check if the actual hook.

MatzeKitt:
Also the check in `wp_footer()` is working for me. Thank you. 👍🏻

inc/class-statify-frontend.php
```diff
@@ -404,7 +408,30 @@ public static function wp_footer() {
 	}
 
 	/**
-	 * Add amp-analytics.
+	 * Add amp-analytics for Standard and Transitional mode.
+	 *
+	 * @see hAMPttps://amp-wp.org/documentation/playbooks/analytics/
```
florianbrinkmann: Here is a little typo: `hAMPttps`

inc/class-statify-frontend.php
```diff
@@ -420,32 +447,35 @@ public static function amp_post_template_analytics( $analytics ) {
...
+				'_ajax_nonce'      => wp_create_nonce( 'statify_track' ),
+				'statify_referrer' => '${documentReferrer}',
+				'statify_target'   => '${canonicalPath}amp/',
+			),
+			'triggers'       => array(
+				'trackPageview' => array(
+					'on'      => 'visible',
+					'request' => 'pageview',
+				),
+			),
```
florianbrinkmann, line 478: As I wrote in Slack, I have no real experience with AMP, but one thing I noticed here: is it correct that the `transport` key was skipped from this config array, or was that accidentally?

inc/class-statify-frontend.php
stklcode: Obviously ... :man_facepalming:

inc/class-statify-frontend.php
stklcode: Good catch! Removing the section was not intentional.

Transport defaults to `beacon`, `xhrpost` and `image` enabled (in hierarchical order), so most browsers will use a compatible method by default. However `image` uses a GET request while _Statify_ requires POST, so I've disabled it explicitly.
Both `beacon` and `xhrpost` perform well.

florianbrinkmann (approved):
Looks good to me

PR #186, opened 2020-09-02
exclude sitemap calls from tracking (#185)

### LINKED ISSUE #185: Sitemap request showing in Statify
I am running XML Sitemap Generator on my blog which is disabling the new core feature from WordPress 5.5 with a filter:
`add_filter( 'wp_sitemaps_enabled', '__return_false' );`

See: https://make.wordpress.org/core/2020/07/22/new-xml-sitemaps-functionality-in-wordpress-5-5/

Although there is no sitemap under `/wp-sitemaps.xml` on my site I have 114 requests tracked in Statify for this address (which results in a 404).

I think we should hide those requests, but unfortunately there is no conditional tag like `is_sitemap()` but there is a comment with an answer on the official post:

> Is there a template function like is_sitemap() or doing_sitemap(). I ask as I have a other filters I running that should not run inside the sitemap.

> There currently isn’t, so checking `get_query_var( ‘sitemap’ )` is probably the easiest way to do that

See: https://make.wordpress.org/core/2020/07/22/new-xml-sitemaps-functionality-in-wordpress-5-5/#comment-39484

stklcode:
resolves #185

The default tracking method does not detect core sitemap calls to _wp-sitemap.xml_, because there is no check. When WP core sitemaps are disabled, the `redirect_template` hook is too early, i.e. the sitemap does not yet know if it shall be generated or skipped, so we accidentally track 404 hits in this case.

JS tracking does not seem to be affected, because there either is no `wo_footer()` for the XML output or `is_404()` returns `true` at this point.

**Solution:**
Add a check for the query variable `sitemap` as suggested in the WP Core discussion<sup>[1](#fn1)</sup> to the `_is_internal()` check, s.t. we never try to track anything for sitemap calls.

----
<a name="fn1">[1]</a> https://make.wordpress.org/core/2020/07/22/new-xml-sitemaps-functionality-in-wordpress-5-5/#comment-39484

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file. This projec
 * Fix AMP compatiblity for Standard and Transitional mode (#181) (#182)
 * JavaScript is no longer embedded if request is served by AMP (#181) (#182)
 * Always register the action for the cleanup (#184)
+* Exclude sitemap calls (WP 5.5) from tracking (#185) (#186)
 
 ## 1.8.0
 * Fix date offset in dashboard widget in WP 5.3+ environments with mixed timezones (#167)
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -252,9 +252,11 @@ class Statify_Frontend extends Statify {
 	 * @return   boolean  $skip_hook  TRUE if NO tracking is desired
 	 */
 	private static function _is_internal() {
-		// Skip for preview, 404 calls, feed, search and favicon access.
+		global $wp_query;
+		// Skip for preview, 404 calls, feed, search, favicon and sitemap access.
 		return is_preview() || is_404() || is_feed() || is_search()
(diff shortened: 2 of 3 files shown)


patrickrobrecht (commented):
Just tested this on WP 5.5 + 5.6 nightly build. I noticed that this PR fixes that `/wp-sitemap.xml` is logged, but I noticed that requests to are still `/wp-sitemap-index.xsl/` counted.

stklcode:
Added another check for the query var `sitemap-stylesheet` which is set for the XSL call. Don't see any more elegant way at the moment, but it works for me.

stklcode:
Rebased on top of the latest development and added test case for sitemaps.

florianbrinkmann (approved):
Looks good to me! Made a few requests to the sitemap file from the `develop` branch, then switched to the PR branch and made more requests. The requests before branch switching were tracked, after switching the branch they were not tracked.
Also ran the tests and everything passed.

PR #187, opened 2020-10-19
Always register the action for the cleanup

### LINKED ISSUE #184: No action shown in WP Crontrol
First reported here: https://wordpress.org/support/topic/no-action-for-statify_cleanup/

I can reproduce this on my website. There is "none" showing for teh action in WP Crontrol for the `statify_cleanup` function scheduled daily. Per documentation from WP Crontrol this could be a bug:

> An event with no action is also useless, although there is a small chance that an action gets conditionally added to this event by a plugin based on a setting, for example, and thee plugin isn’t correctly handling the scheduling or unscheduling of the event.

On the other side BackWPup and Antispam Bee have the same warning ...

Maybe because we add this in a private function?

patrickrobrecht:
Resolves #184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. This projec
 ## 1.8.1
 * Fix AMP compatiblity for Standard and Transitional mode (#181) (#182)
 * JavaScript is no longer embedded if request is served by AMP (#181) (#182)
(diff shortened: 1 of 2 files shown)


stklcode (approved):
Would be enough to register it for cron and admin calls. But the overhead of registering an unused action is minimal, so I’m fine with the proposed solution.

PR #190, opened 2020-12-09
release 1.8.1

stklcode:


readme.txt
```diff
@@ -116,6 +117,13 @@ has to be added to the theme's `functions.php`. The condition has modified such
 ## Changelog ##
 You can find the full changelog in [our GitHub repository](https://github.com/pluginkollektiv/statify/blob/master/CHANGELOG.md).
 
+### 1.8.1
+* Fix AMP compatibility for Standard and Transitional mode (#181) (#182)
+* JavaScript is no longer embedded if request is served by AMP (#181) (#182)
+* Always register the action for the cleanup (#184)
+* Exclude sitemap calls (WP 5.5) from tracking (#185) (#186)
```
florianbrinkmann: The »(WP 5.5)« might sound like that is something only for WP 5.5, maybe change it to something like »Exclude calls to the sitemap added in WP 5.5 from tracking (#185) (#186)« or so?

florianbrinkmann (approved):
Looks good to me.
Made a suggestion for a small change in the changelog, but maybe it is just me and it is clear as it is :)

readme.txt
stklcode: We used `WP 5.3+` in the 1.8.0 changelog. I'd be fine with either your suggestion or the shorthand "(... WP5.5+)".

florianbrinkmann:
»WP5.5+« is good!

patrickrobrecht (approved):

PR #191, opened 2020-12-12
update test environment

stklcode:
PHPUnit 7.5 has sufficient support for PHP 8, even though it is not compatible officially, and WP test does not yet support PHPUnit 8 or later. PHPUnit 5/6 however fail with PHP 8, so we use a different dev-dependency set for PHP 5 integration tests.

With the proposed changes we are running 3 test cycles:
* WP 4.7 with PHP 5.6 (oldest supported WP with oldest feasible PHP)
* WP 5.5 with PHP 7.4 (latest combination supported without platform overrides in Composer)
* WP 5.6 with PHP 8.0 (latest WP and PHP)

Dropped PHP 7.0-7.3 from CI roster, as there is no real benefit in running PHPCS lints with multiple PHP versions, i.e. we only test the environment in these jobs, not actually our code.

Also updated Node environment to Node 14 and latest dev-dependencies.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -17,9 +17,11 @@
 /.stylelintrc.json export-ignore
 /.travis.yml export-ignore
 /composer.json export-ignore
+/composer-php5.json export-ignore
 /package.json export-ignore
 /package-lock.json export-ignore
 /composer.lock export-ignore
+/composer-php5.lock export-ignore
 /phpcs.xml export-ignore
 /phpunit.xml export-ignore
 /README.md export-ignore
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
 
 before_script:
   - composer install --ignore-platform-reqs
-  - npm install
+  - npm ci
   - |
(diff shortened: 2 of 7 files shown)


patrickrobrecht (commented):
In #190 the WP version for the test with PHP 7.4 was updated to use the WP 5.6 as well. If the merge conflict is resolved, this PR  LGTM. In the long term we should decide how long Statify continues to support PHP 5. 

Do we need to keep the `php: nightly` which would be still running without the actual tests?

stklcode:
> Do we need to keep the php: nightly which would be still running without the actual tests?

I don't think so.
Pure unit tests against _nightly_ are a good indicator if our code does support the upcoming PHP release, but with tests against an actual WP instance the jobs have several reasons to fail. So the builds don't tell any story unless we have a closer look into what has failed.
PHPCS-only for nightly only tells us if devenv is compatible ... but with ` --ignore-platform-reqs`  is only the PHPCS runtime which is "tested".

patrickrobrecht (approved):

PR #195, opened 2020-12-14
add "?amp" instead of "/amp/" for AMP visit tracking (#183)

### LINKED ISSUE #183: Wrong path for AMP pages
Currently, for AMP pages the URL gets extended by an `amp/` (https://github.com/pluginkollektiv/statify/blob/develop/inc/class-statify-frontend.php#L428). However, this will result in wrong URLs since https://example.com/ will become https://example.com/amp/ in the database. The correct way would be to add an `?amp` or `&amp` to the URL to get a valid URL, which is the AMP variant of the page.

stklcode:
This is one possible way to fix #183 

Instead of adding ` /amp/`  to the URL of the page is accessed via AMP, we now add ` ?amp` .
This is always fine, because AMP should never provide other URL parameters.

Depending on the outcome of the discussion in #183 we can either merge this PR or close it. Just opened, so we don't do the work twice if needed.

### DIFF
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -163,7 +163,7 @@ class Statify_Frontend extends Statify {
 			),
 			'extraUrlParams' => array(
 				'referrer' => '${documentReferrer}',
-				'target'   => '${canonicalPath}amp/',
+				'target'   => '${canonicalPath}?amp',
 			),
 			'triggers'       => array(
 				'trackPageview' => array(
--- a/inc/class-statify.php
+++ b/inc/class-statify.php
@@ -135,9 +135,15 @@ class Statify {
 		/* Global vars */
 		global $wp_rewrite;
 
-		// Trim target URL.
+		// Trim target URL, i.e. remove query parameters..
 		if ( $wp_rewrite->permalink_structure ) {
-			$target = wp_parse_url( $target, PHP_URL_PATH );
+			$parsed_target = wp_parse_url( $target );
+			$target = isset( $parsed_target['path'] ) ? $parsed_target['path'] : null;
+
+			// Re-add AMP parameter to preserve that information (only applicable for JS tracking).
+			if ( isset( $parsed_target['query'] ) && 'amp/' === $parsed_target['query'] ) {
+				add_query_arg( 'amp', '', $target );
+			}
 		}
 
 		// Init rows.


inc/class-statify-frontend.php
```diff
@@ -91,15 +91,24 @@ public static function track_visit( $is_snippet = false ) {
...
 		if ( $wp_rewrite->permalink_structure ) {
-			$data['target'] = wp_parse_url( $data['target'], PHP_URL_PATH );
+			$parsed_target  = wp_parse_url( $data['target'] );
+			$data['target'] = $parsed_target['path'];
+
+			/*
+			 * Re-add AMP parameter to keep that information (only applicable for JS tracking).
+			 * We can assume "amp" to be the only parameter, as ${canonicalPath} already eliminates other queries.
+			 */
+			if ( $is_snippet && isset( $parsed_target['query'] ) && 'amp/' === $parsed_target['query'] ) {
```
pfefferle: what if `amp` is not the last query param?

inc/class-statify-frontend.php
```diff
@@ -91,15 +91,24 @@ public static function track_visit( $is_snippet = false ) {
...
+				$data['target'] .= '?amp';
```
pfefferle: What about using `add_query_arg` even if the query string should be empty?

inc/class-statify-frontend.php
```diff
@@ -469,7 +478,7 @@ private static function make_amp_config() {
 				'action'           => 'statify_track',
 				'_ajax_nonce'      => wp_create_nonce( 'statify_track' ),
 				'statify_referrer' => '${documentReferrer}',
-				'statify_target'   => '${canonicalPath}amp/',
+				'statify_target'   => '${canonicalPath}?amp',
```
pfefferle: Same as above, why not use `add_query_arg`?

inc/class-statify-frontend.php
```diff
@@ -91,15 +91,24 @@ public static function track_visit( $is_snippet = false ) {
...
-		/* Relative target url */
+		// Relative target URL.
 		$data['target'] = user_trailingslashit( str_replace( home_url( '/', 'relative' ), '/', $target ) );
 
-		// Trim target url.
+		// Trim target URL, i.e. remove query parameters.
 		if ( $wp_rewrite->permalink_structure ) {
-			$data['target'] = wp_parse_url( $data['target'], PHP_URL_PATH );
+			$parsed_target  = wp_parse_url( $data['target'] );
+			$data['target'] = $parsed_target['path'];
```
pfefferle: can it be possible that `$parsed_target['path']` is not set?

inc/class-statify-frontend.php
stklcode: What's the benefit at this point?
`${canonicalPath}` is passed as is and will be replaced in the frontend logic by Javascript.

`add_query_arg(array('amp' => ''), '${canonicalPath}')` will always return `${canonicalPath}?amp`.

I'm not 100% sure if there is any way to get a valid query part in the canonical path, at least I've never seen this.

Finally this is not even a real query parameter, rather than a marker. We detect the `?amp` suffix, no matter if this is even a valid URI (we strip it off and re-add it for further processing)

inc/class-statify-frontend.php
stklcode: Good catch, we should add validation and pass `null` if not filled (at least that's what `wp_parse_url('https://example.com', PHP_URL_PATH)`  would have returned)

inc/class-statify-frontend.php
pfefferle: Haven't read the code comment :(

stklcode: Strictly speaking it's not a real query parameter at all, just a suffix we add. `${canonicalPath}?amp` will never result in `?amp` not being the very last 4 characters of the argument.

That's not really nice though. We probably should make this clear and split it off before parsing the URL and check against `str_ends_with( $data['target'], '?amp' )` (or equivalent legacy PHP variant)

inc/class-statify-frontend.php
stklcode: Using `add_query_arg( 'amp', '', $data['target'] )` would not hurt at this point.

As far as I remember query arguments should never be part of canonical paths, but we never know.

inc/class-statify-frontend.php
pfefferle: You are right, I was not aware of the frontend logic!

PR #198, opened 2021-03-10
Use full width for total to avoid line break

patrickrobrecht:
Avoids the line break for long dates ("since Monday, 7. September 2020" as in the screenshot of #197) in the dashboard.


stklcode:
Didn’t verify, but isn’t the `float: left` obsolete in this case, too?

Would prefer a simple `flex` solution in general, but the the minimal change is totally fine here.

patrickrobrecht:
Closed in favor of #199.

PR #206, opened 2021-05-25
remove double closing tag on value attribute of radio buttons

stklcode:
Radio buttons in the settings form, i.e. selector for JS tracking mode contains a superfluous `>` character in the value attribute because of doubled closing tag after the PHP sequence.
```html
<input name="statify[snippet]" type="radio" value="0>">
                                                    ^
----------------------------------------------------|
```


florianbrinkmann (approved):
LGTM

PR #207, opened 2021-05-27
introduce SonarCloud analysis and GH actions

stklcode:


patrickrobrecht:
+1 for replacing Travis with a GitHub action. We need another job for the JavaScript tests (#205).

I don't like the current setup for the PHP 5.6 tests (separate `composer.json`) - if that's the point preventing the switch to GitHub, I'm fine with removing them and removing PHP 5 support at all. Currently we even claim backwards compatibility with PHP 5.2

stklcode:
> I don't like the current setup for the PHP 5.6 tests (separate composer.json) - if that's the point preventing the switch to GitHub

That's legacy behavior copied from the Travis setup.

https://github.com/pluginkollektiv/statify/blob/eade6c8ad7b68024f1d7bcaff759d752c07b83a2/.travis.yml#L8-L11

There are 2 possible ways to get around this:
* remove the _composer.lock_
  We do have non-dev-dependencies (Chartist) here. Unlikely to break however.
* remove the _composer.lock_ in the PHP 5 CI job or run `composer update` instead of `install` 

Currently does not work due to incompatibility between the PHP and MySQL versions in the runner, but that's a different story.

> I'm fine with removing them and removing PHP 5 support at all.

Sure we can remove the support, but that should be addressed in a different issue/PR.

stklcode:
One thing we should discuss about the analysis. There are two approaches
* automatic analysis, statically without coverage
  only requires a _.sonarcloud.properties_ file, the rest is done automatically by the GH integration
* active analysis from CI
  actively run the sonar scanner after the tests steps as final part of the GH action
  
  Currently automatic analysis is enabled and because of that the analysis step fails:
> You are running CI analysis while Automatic Analysis is enabled. Please consider disabling one or the other.

If we decide against CI analysis, we must remove the steps from the pipeline.

stklcode:
> Currently does not work due to incompatibility between the PHP and MySQL versions in the runner, but that's a different story.

Fixed using Ubuntu 18.04 runners that ship MySQL 5.7 instead of 8, so the old PHP mysqli is able to login in default configuration.

stklcode:
Reverted back to static automatic analysis, to get this done soon without major impacts.

florianbrinkmann (approved):

PR #209, opened 2021-05-28
fix CSS to not overrule display property on collapsed widget

### LINKED ISSUE #208: Statify widget not collapsed
WordPress offers to collapse widgets: It adds an `closed` class to the widget `<div>`.

Our CSS for `#statify_dashboard .inside` uses `display: flex` which overrides the `display: none;` for `.js .closed .inside` from WordPress Core.

stklcode:
resolves #208 

We could have used the shorthand selector `#statify_dashboard:not(.closed) .inside`, but this would be incompatible with the IE browser and we still claim to support WP versions that have not dropped the support yet. So we simply add another directive to overrule the selector.

Refactoring the stylesheet to not use IDs at all would be great, but for now I'd prefer a quick non-invasive fix.


florianbrinkmann (approved):
LGTM

PR #210, opened 2021-06-20
Declares multisite init_site methods as static

walterebert:
In a multisite setup on PHP 8 initialisation methods cause a fatal error when adding/deleting a site. Declaring these methods as static fixes the problem.

`Fatal error: Uncaught TypeError: call_user_func_array(): Argument #1 ($function) must be a valid callback, non-static method Statify_Install::init_site() cannot be called statically`

### DIFF
--- a/inc/class-statify-install.php
+++ b/inc/class-statify-install.php
@@ -53,7 +53,7 @@ class Statify_Install {
 	 *
 	 * @param int $site_id Site ID.
 	 */
-	public function init_site( $site_id ) {
+	public static function init_site( $site_id ) {
 
 		switch_to_blog( (int) $site_id );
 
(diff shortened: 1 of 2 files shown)


stklcode:
Thanks @walterebert for this fix.

The non-static declaration was probably never correct and did raise a warning with the latest PHP 7.x versions. Solutions looks reasonable.

Automated tests are not running for another reason, I will have a closer look at it later today.

walterebert:
Unit tests for WordPress don't work with PHP 8 as far as I know. PHPUnit Polyfills might help, but I haven't used it yet: https://github.com/Yoast/PHPUnit-Polyfills

stklcode:
I'm aware of the restrictions due to the PHPUnit 7 limit and our test cases do run fine with PHP 8 as long as you install with ` --ignore-platform-req=php`. Current reason is the _travis-ci.org_ to _travis-ci.com_ migration that has not been done yet for this project, because we actually wanted to move to GH actions - the corresponding PR has not been merged either, so there's no CI at the moment.

stklcode (approved):
LGTM

PR #213, opened 2021-07-17
ci: unify composer setup for CI workflows

stklcode:
The Composer action is fine, but always uses latest PHP, which requires some compatibility flags and generates log files in the source directory that must not be added to the deployment bundle.

We now use the same stack in test and deployment actions which solves both issues.


patrickrobrecht (approved):
LGTM.

PR #214, opened 2021-07-17
ci: run tests against "latest" and "nightly" WP version

stklcode:
Keep older stacks like PHP/WP 5.6/4.7 and 7.4/5.5 as fixed marks, but replace the 8.0/5.8 with 8.0/latest.

Also add 8.0/nightly step which might possibly fail, but gives a hint on compatibility with upcoming releases. The latter should not be added as mandatory check though.


patrickrobrecht (approved):
LGTM.

PR #226, opened 2022-02-19
migrate to WP API (#219)

### LINKED ISSUE #219: Statify REST API Endpoint
This issue is meant to be a starting point for discussing the usage of a custom REST API endpoint for tracking and/or showing data.

stklcode:
This PR addresses #219

### Introduce API endpoints

We introduce 2 API routes to model current behaviors:

#### Tracking

POST `/statify/v1/track`
Track a visit via JS. This is intended to replace the WP AJAX calls.

We override the default authentication mechanism ([ref](https://developer.wordpress.org/rest-api/using-the-rest-api/authentication/) for the track route. This would unauthenticate every request, i.e. `is_user_logged_in()` will always return `false`, if no nonce is present and returns 403 if the given nonce is expires.

So we do not send any nonce at all, if verification is disabled, and verify the nonce from request parameters manually, if necessary.

#### Stats retrieval

GET `/statify/v1/stats`
Retrieve statistics data. Is uses the same authentication requirements as before and returns dashboard data as JSON:
```json
{
  "visits": {
    "2022-02-18": 17,
    "2022-02-19": 4
  },
  "referrer": [
    { "host": "example.com", "url": "https://example.com/test/", "count": 3 }
  ],
  "target": [
    { "url": "/testpage/", "count": 10 }
  ],
  "totals": {
    "today": 4,
    "alltime": 21,
    "since": "2022-02-18"
  }
}
```
### Tracking changes

Javascript and AMP tracking issue POST requests to the new API endpoint instead of using _admin-ajax.php_

### Dashboard widget changes

The dashboard widget is no longer privisioned from a hidden table. The backend just generated a skeleton with placeholders that will be replaced after the data is retrieved from the API.

statify_widget_placeholder


stklcode:
The "since 1.9" is not necessarily correct. I'd say this is quite a major change, so we should probably name it 2.0 if it's getting merged.

We switched from some custom logic to WP AJAX in 1.7 though and one might say the "tracking mechanism itself is not designed as public API.

florianbrinkmann:
I know we cannot use many features of modern WP without JS, but I wonder if we couldn’t load the initial data via PHP and not JS?

stklcode:
Sure we could.

Top lists could be rendered inline instantly. But charts are non-blocking JS-generated anyway, so there is always at least a minimal delay in the UI, even if data is read from hidden markup like before. It would just introduce some (rather small) additional JS code to have two different ways to fetch data.

Main benefit of dynamic requests I see here is that we could skip the data completely, if the widget is initially collapsed and fetch it when it’s actually opened (i.e. add a simple listener to the expand icon). So there is a chance to reduce overhead when accessing the dashboard. Same trade-off as always.

florianbrinkmann:
Oh, yes, didn’t think about the chart and that it requires JS 🙈 Will dig deeper into the PR and give feedback :)

inc/class-statify-dashboard.php
```diff
@@ -129,22 +125,25 @@ public static function add_js() {
...
 		wp_localize_script(
 			'statify_chart_js',
-			'statify_translations',
+			'statifyDashboard',
 			array(
-				'pageview'  => strip_tags( esc_html__( 'Pageview', 'statify' ) ),
-				'pageviews' => strip_tags( esc_html__( 'Pageviews', 'statify' ) ),
+				'i18n'  => array(
+					'error'        => strip_tags( esc_html__( 'Error loading data.', 'statify' ) ),
+					'nodata'       => strip_tags( esc_html__( 'No data available.', 'statify' ) ),
```
florianbrinkmann: Doesn’t `strip_tags` and `esc_html` do the same here?

inc/class-statify-api.php
```diff
@@ -0,0 +1,152 @@
...
+
+		add_filter( 'rest_authentication_errors', array( __CLASS__, 'check_authentication' ), 5 );
+	}
+
+	/**
+	 * Filters REST API authentication errors.
+	 *
+	 * Override the default authentication check for the tracking endpoint.
+	 * Requests are unauthenticated by default, if no nonce is provided. Verification can be disabled in the plugin
+	 * configuration, and we still need the information, if tue current usr is logged in.
```
florianbrinkmann: Small typo here, `usr`

florianbrinkmann:
Had two small notes/questions, besides that LGTM

inc/class-statify-dashboard.php
florianbrinkmann: No, doesn’t, but after the HTML was escaped, the `strip_tags` shouldn’t find something

stklcode: Absolutely right. We should remove the `strip_tags` and stick with the commonly used `esc_html` here.

Just moved the block to the `i18n` subarray, but if we modify this code anyway, seems like a good point to clean this up.

stklcode:
Resolved both and rebased the branch. Think we should not squash tracking and dashboard changes together, so I squashed both changes into the corresponding commits.

As previously said, we should think about a target version. 

We did a AJAX migration back in 1.7.0. IIRC we considered the routine being internal logic, even though it apparently was kind of breaking for some users who had AJAX access blocked or long caching times.
I know there are some folks out there that trigger tracking differently, so my preference would be 2.0 here and not 1.x. (even though the inline docs state `@since 1.9` now - but sanity checks on these annotation should be part of release work anyway, so might be postponed).

florianbrinkmann (approved):
LGTM. For me it looks like a breaking change, too, so would vote for a v2

stklcode:
Rebased onto latest _develop_ state:
* updated `@since` tags to 2.0.0
* merged conflicts from modified bot detection
* added issue and PR id to commit messages

Issue #78, opened 2018-03-26
Show title instead of permalink


Zodiac1978:
Use post/page title instead of permalinks. Needs some more research for edge cases like archive pages etc.

stklcode:
Pipe the target through WP query and and extract Post title might already cut it for 95%.

Everything else is also there, if we process the complete target query string, "just" a matter of how to display.

We should also limit this to the top targets when generating the list or at least on some aggregation level as the overhead is far too heavy to do this for every single row in the stats table 😉

websupporter:
`url_to_postid()` would be my first choice, if it would be only about posts/pages.

If it comes to archive pages, I want to draw attention to those titles:
https://de.wordpress.org/category/veranstaltungen/page/3/
https://de.wordpress.org/category/veranstaltungen/page/2/

It you would only display the title, the user would have no possibility to distinguish between those as the current page is not part of the `<title>`-element.

patrickrobrecht:
In _Statify - Extended Evalution_ I use:
```
if ( '/' === $url ) {
	return __( 'Home Page', 'extended-evaluation-for-statify' );
}
$post_id = url_to_postid( $url );
return ( 0 === $post_id ) ? esc_url( $url ) : get_the_title( $post_id )
```

This leads to a better output for many posts (at least better than the current permalink). There will remain some cases which are not handled by this, e.g. category/tag pages as mentioned by @websupporter already, deleted posts whose URL can still be in the Statify table.

mahype:
I have added a very first version of url > title replacement. Now we replacing the title of

- Home Page
- Post Types
- Categories (and only post categories!)

We have only categories for now, because I couldn't find a function similar to get_category_by_path for terms in general. All the code for categories which could be used for other terms is hard coded for categories at the moment. You can see this here:

https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/category.php#L126

There are three possibilities now:

1. Writing a functionn which determines term names.
2. Usinng anoter function if somebody else knows one.
3. Writing a change/pull request for WordPress to abstract the functionality direct in WordPress.

PR #230, opened 2022-07-07
Widget title addition

John-H-Smith:
Added the title to the widget on hover over url so that you can see what page it is if you have short urls enabled.

image

If it's the front page, the title of the website is returned.


stklcode:
related to #78

pfefferle (approved):
LGTM

2ndkauboy:
It seems that the `url_to_postid()` function (which is also quite heavy) does not really work for many types of URLs, such as CPT single views, CPT archives and even category archives. I'm also a bit unsure if this is a good idea in terms of a11y. :thinking:

John-H-Smith:
> It seems that the `url_to_postid()` function (which is also quite heavy) does not really work for many types of URLs, such as CPT single views, CPT archives and even category archives. I'm also a bit unsure if this is a good idea in terms of a11y. thinking

Yeah, I thought about that and verified the behaviour with a category archieve. It seems like the function _[get_page_by_path](https://developer.wordpress.org/reference/functions/get_page_by_path/)_ could be working, but it seems to be quite buggy when switching the permalink structure, so I had switched to _url_to_postid_. Do you have another function which could be working too?

views/widget-front.php
```diff
@@ -71,7 +71,7 @@ class_exists( 'Statify' ) || exit;
 							<?php echo (int) $target['count']; ?>
 						</td>
 						<td class="t">
-							<a href="<?php echo esc_url( home_url( $target['url'] ) ); ?>" target="_blank" rel="noopener noreferrer">
+						<a href="<?php echo esc_url( home_url( $target['url'] ) ); ?>" target="_blank" rel="noopener noreferrer" title="<?php echo 0 == url_to_postid( $target['url'] ) ? esc_html( get_bloginfo( 'name' ) ) : esc_html( get_the_title( url_to_postid( $target['url'] ) ) ); ?>">
```
stklcode: [`esc_html`](https://developer.wordpress.org/reference/functions/esc_html/) escapes HTML code like `<`, `>` and some special characters.

You are populatione the `title` _attribute_ here, so slightly different escaping is required here, like quotes `"`.
E.g. `foo=" bar="test` passes HTML escaping, but the final code would be `title="foo" bar="test"` - obviously not quite the desired result.

We should use [`esc_attr`](https://developer.wordpress.org/reference/functions/esc_attr/) in this case.

stklcode (changes requested):

MatzeKitt:
I think the cleanest way would be in storing metadata for the visited page in the Statify table, like post ID as an object ID and the object type, which could be any post type or term type. That way you can find the correct object according to object type and object ID and get the current title of the object.

John-H-Smith:
> I think the cleanest way would be in storing metadata for the visited page in the Statify table, like post ID as an object ID and the object type, which could be any post type or term type. That way you can find the correct object according to object type and object ID and get the current title of the object.

Yeah, that would be. But as of that, you would have to update the metadata every time the url or the title is updated - I think thats a quite heavy performance issue.

MatzeKitt:
> But as of that, you would have to update the metadata every time the url or the title is updated - I think thats a quite heavy performance issue.

If you construct it correctly, you can retrieve the current title of all items in the widget with a single SQL query, which is much better than using `url_to_postid` for every single item.

John-H-Smith:
> If you construct it correctly, you can retrieve the current title of all items in the widget with a single SQL query, which is much better than using `url_to_postid` for every single item.

Yeah, that's correct, it would be much better. But modifying the table after it has been created would force you to either uninstall and reinstall statify to create the table correctly or check for every website access if the new columns in the table already exist.
Also, I just noticed that the dashboard widget limits the count of all items to 100, so I am pretty sure using `url_to_postid` would be no problem.

stklcode:
> But modifying the table after it has been created would force you to either uninstall and reinstall statify to create the table correctly [...]

Actually that's not a big deal at all. There is already a routine to initialize the table properly, so the schema may be updated during a plugin update.

MatzeKitt:
As addition: take a look into [`dbDelta`](https://developer.wordpress.org/reference/functions/dbdelta/)

John-H-Smith:
Alright - but with what action hook should I modify the table in your opinion? Directly in the functions file for execution on every call would be a bad practice I think...

MatzeKitt:
I can only speak for myself, but the best way I found was to store the migration version as option and migrate the database depending on this option.

See here: https://github.com/epiphyt/embed-privacy/blob/main/inc/class-migration.php

PR #233, opened 2022-08-26
update test framework to  PHPUnit 9 with polyfills

stklcode:
Development and testing of Statify becomes quite a pain on modern
systems as compatibility gaps between PHP, PHPUnit and WP increase.

With WP 5.9 we got some handy polyfills for extended PHPUnit
compatibility, s.t. we finally can use PHPUnit 9 while PHPUnit 7 won't
even initialize on PHP 8.1

We add some dirty hacking here to leverage the polyfills with legacy WP
4.7 setup to preserve a case for the oldest supported version.

Update test matrix up to PHP 8.2 and WP 6.1.

Finally, we move to "ubuntu-latest" runners with dedicated MySQL setup
action instead of deprecated 18.04 runners.


composer.json
```diff
@@ -10,8 +10,8 @@
   ],
   "require": {
     "php": "^5.2|^7|^8",
-    "npm-asset/chartist": "^0.11.4",
-    "npm-asset/chartist-plugin-tooltips-updated": "^0.1.4"
```
florianbrinkmann, line 14: Why the removal of the `^` here and in the previous line?

florianbrinkmann (approved):
LGTM

composer.json
stklcode: We previously had version `^0.11.4` and a _composer.lock_ file that locked the version to `0.11.4` when installing. We now remove the lockfile for convenience (we otherwise would have to keep at least 3 locks for different PHP generations), so Composer would pick up the latest available version `0.11.4 <= v < 1`, so we lock it this way.

Doesn't make any difference for this specific dependency, because 0.11.4 is effectively the latest version below 1 and the Update is already in place (#239).

florianbrinkmann: ah okay, thanks!

PR #234, opened 2022-09-08
introduce data aggregation (#194)

### LINKED ISSUE #194: Data aggregation
Multiple rows with the same date, target and referrer can be aggregated with an additional column `count`.

- Database migration (add column `count`, migrate existing data)
- In the queries `count(*)` needs to be replaced with `SUM(count)`.
- When adding a new row, check whether a row with that date, target and referrer exists already. If it exists, increase `count`, otherwise create it with `count = 1`.

stklcode:
We add a new column named `hits` to the database table with default value of 1.
The tracking routine is preserved as is, i.e. it will insert a single row per hit.

The cleanup routine (cron job) is extended with an aggregation function that walks through the database and groups distinct date/referrer/target combinations with the sum of all hits.

*before:*
| created    | referrer            | target | hits |
| ---------- | ------------------- | ------ | ---- |
| 2022-09-07 | example.com         | /      | 1    |
| 2022-09-07 | pluginkollektiv.org | /test/ | 1    |
| 2022-09-07 | example.com         | /      | 1    |
| 2022-09-07 | example.com         | /      | 1    |
| 2022-09-08 | pluginkollektiv.org | /test/ | 1    |
| 2022-09-08 | pluginkollektiv.org | /      | 1    |

*after:*
| created    | referrer            | target | hits |
| ---------- | ------------------- | ------ | ---- |
| 2022-09-07 | example.com         | /      | 3    |
| 2022-09-07 | pluginkollektiv.org | /test/ | 1    |
| 2022-09-08 | pluginkollektiv.org | /test/ | 1    |
| 2022-09-08 | pluginkollektiv.org | /      | 1    |

The output is generated from `SUM(hits)` instead of `COUNT(*)` then.

We also introduce a boolean book `statify__skip_aggregation` (defaults to `false`) to keep the previous behavior and disable the new aggregation.

This might be necessary if the inserted entry is extended with custom columns, e.g. in a `statify__visit_saved` hook. If this is the case, aggregation will fail, because we don't know about additional columns. We could of course add another hook for custom aggregation columns and dynamically build up the statements, but for most users this won't be necessary and it can be extended in future releases as well.

### DIFF
--- a/inc/class-statify-cron.php
+++ b/inc/class-statify-cron.php
@@ -23,13 +23,12 @@ class Statify_Cron extends Statify {
 	 *
 	 * @since    0.3.0
 	 * @version  1.4.0
+	 * @wp-hook boolean  statify__skip_aggregation
 	 */
 	public static function cleanup_data() {
-
-		// Global.
 		global $wpdb;
 
-		// Remove items.
+		// Remove old items.
 		$wpdb->query(
 			$wpdb->prepare(
 				"DELETE FROM `$wpdb->statify` WHERE created <= SUBDATE(%s, %d)",
@@ -38,13 +37,13 @@ class Statify_Cron extends Statify {
 			)
 		);
 
-		// Aggregate.
-		self::aggregate_data();
+		// Aggregate data.
+		if ( ! apply_filters( 'statify__skip_aggregation', false ) ) {
+			self::aggregate_data();
+		}
 
 		// Optimize DB.
-		$wpdb->query(
-			"OPTIMIZE TABLE `$wpdb->statify`"
-		);
+		$wpdb->query( "OPTIMIZE TABLE `$wpdb->statify`" );
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


inc/class-statify-frontend.php
```diff
@@ -79,6 +79,7 @@ public static function track_visit( $is_snippet = false ) {
 			'created'  => '',
 			'referrer' => '',
 			'target'   => '',
+			'hist'     => 1,
```
MatzeKitt: typo?
`hist` => `hits`

stklcode: yep 🙈

patrickrobrecht:
These changes would require changes to keep the functionality of Statify - Extended Evaluation, but hopefully helps to improve the  performance with large Statify tables (I've installs with 1,000,000+ entries): https://github.com/patrickrobrecht/extended-evaluation-for-statify/issues/29.

stklcode:
As previously discussed in various places this feature might need some additional planning.

* With planned additional data collection (probably even dynamically extensible) this might require constant refactoring.
* Is database size really still an issue so that it's worth the effort?
* Probably drop this idea in favor of maybe some more elaborate caching of evaluation.

PR #235, opened 2022-11-03
clean up HTML markup for dashboard widget

stklcode:
* remove redundant for/id attributes from wrapped `<label>` tags
  The `<input` elements are are wrapped `<label>`, so the association is clear and no for/id mapping is required.
* use _<div>_ instead of _<fieldset>_ as wrapper around settings
  For valid HTML the `<fieldset>` should have a `<legend>` element. We might move the headline (`<h3>`) into the legend, but actually we only have one set of fields in the form, so additional semantic grouping is not necessary.
* do not close <input> tags
  We close the tags for number inputs, but not for checkboxes. Let's use the same way for all inputs.
* add type attribute to refresh button
  Type attribute is required. All browsers tested seem to assume "submit" by default (otherwise the mechanism would not work), but we should make it explicit.

### DIFF
--- a/views/widget-back.php
+++ b/views/widget-back.php
@@ -22,27 +22,27 @@ class_exists( 'Statify' ) || exit; ?>
 <?php endif; ?>
 
 <h3><?php esc_html_e( 'Widget Settings', 'statify' ); ?></h3>
-<fieldset>
-	<label for="statify_days_show">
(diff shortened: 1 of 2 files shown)


Zodiac1978 (approved):
LGTM!

PR #236, opened 2022-11-04
extend options to allow skipping only administrators (#229)

### LINKED ISSUE #229: Add option to exclude administrators
There already is an option to exclude logged in users. However, in environments with e.g. shop systems customers are logged in. So currently you need to decide whether you wand to exclude all logged in users, and thus your customers, too, or none. However, I don’t want to be included as administrator. Thus I would love to have a separate option for that.

Or alternatively, add a multi select with all available roles so that you can pick all roles explicitly, which you want to have excluded.

stklcode:
resolves #229 

The current option to skip logged-in users is extended with another possible value to exclude administrators (users who can manage options), but track regular users' visits.

The checkbox option is extended to a select-box with 3 possible options. Current internal config structure (0 = track users, 1 = exclude users) remains fully compatible while new value (2) represents "skip administrators".

### DIFF
--- a/inc/class-statify-settings.php
+++ b/inc/class-statify-settings.php
@@ -323,9 +323,13 @@ class Statify_Settings {
 	 */
 	public static function options_skip_logged_in() {
 		?>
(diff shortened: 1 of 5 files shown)


stklcode:
Resolved conflicts in test cases with AJAX-to-API migration.

florianbrinkmann (approved):
Tested and works like expected.

PR #237, opened 2022-11-04
replace FILTER_SANITIZE_STRING with sanitize_text_field

stklcode:
The string filter is deprecated as of PHP 8.1, so let's use another appropriate way to sanitize the input.


Zodiac1978 (approved):
LGTM!

PR #238, opened 2022-11-04
prepare release 1.8.4

stklcode:
* add relevant entries to changelog
* raise version to 1.8.4 in _statify.php_
* raise stable version to 1.8.4 in _readme.txt_
* declare compatibility with WordPress 6.1 in _readme.txt_

Compatibility should be tested one last time, then we should be good to go for the maintenance release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
 
-## unreleased
+## 1.8.4
 * Use same date retrieval for tracking and analysis (#227) (#232)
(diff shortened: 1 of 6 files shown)


stklcode:
Tested on a WP 6.0 / PHP 8.0 and WP 6.1 / PHP 8.1 site both WP_DEBUG enabled without any issues.

Zodiac1978 (changes requested):
Little typo "abd" -> "and" in upgrade notice of readme.txt

It seems you forgot to raise the version number in the statify.php plugin header

stklcode:
And that's why Friday-evening-changes always benefit from a second pair of eyes :see_no_evil: 

Should be fixed now.

Zodiac1978 (approved):
LGTM!

PR #239, opened 2022-12-15
update Chartist library to v1.3

stklcode:
Chartist library has been target of major refactoring. Update old v0.11 to latest v1.3 and the patched tooltip plugin v1.0.

~The tooltip plugin works fine with the new Chartist, if the names of the chart classes are updated. We apply 2 hacks to make it compatible:~
 * ~alias the Chartist version in the Composer file~
 * ~alias the Bar/Pie/LineChart classes in our wrapper script~

----

~This little hack seems acceptable to me and not as disrupting as e.g. shadowing the tooltip plugin in our sources.~

~A patch for the plugin library has been submitted upstream (https://github.com/LukBukkit/chartist-plugin-tooltip/pull/20), which basically proposes renaming, but I don't have any idea if and when this will be published.~

(update: the patch was merged in the plugin v1.0.0, so we don't need custom patches anymore)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,8 +10,8 @@
   ],
   "require": {
     "php": "^5.2|^7|^8",
-    "npm-asset/chartist": "0.11.4",
-    "npm-asset/chartist-plugin-tooltips-updated": "0.1.4"
+    "npm-asset/chartist": "^1.3.0",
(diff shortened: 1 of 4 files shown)


florianbrinkmann (approved):
LGTM

PR #242, opened 2023-03-18
Track internal search queries

### LINKED ISSUE #220: Count internal search queries
What about tracking internal search terms?

florianbrinkmann:
This PR:
* ensures search queries are stored in the DB (lowercased, and never with a trailing slash)
* Adds a `Statify_Dashboard::parse_target()` method that is called for the targets list in the widget. It makes sure the search query items are not just displayed as `/?s=[…]`

Looks like this:

Bildschirm­foto 2023-03-18 um 14 09 48

Closes #220


inc/class-statify-frontend.php
```diff
@@ -92,11 +92,23 @@ public static function track_visit( $is_snippet = false ) {
...
-			$data['target'] = wp_parse_url( $data['target'], PHP_URL_PATH );
+			if ( 0 !== strpos( $data['target'], '/?s=' ) ) {
+				// No search, so we only need the `PHP_URL_PATH`.
+				$data['target'] = wp_parse_url( $data['target'], PHP_URL_PATH );
+			}
+		}
+
+		// Convert search query to lowercase.
+		if ( 0 === strpos( $data['target'], '/?s=' ) ) {
+			$data['target'] = strtolower( $data['target'] );
```
stklcode: We could think of slightly refactoring this routine to not repeat the search-check thrice in a row.

Something like this should do the same
```php
if ( 0 !== strpos( $data['target'], '/?s=' ) ) {
	// Add trailing slash if that is no search query.
	$data['target'] = user_trailingslashit( $data['target'] );

	// Trim target url.
	if ( $wp_rewrite->permalink_structure ) ) {
		// No search, so we only need the `PHP_URL_PATH`.
		$data['target'] = wp_parse_url( $data['target'], PHP_URL_PATH );
	}
} else {
	// Convert search query to lowercase.
	$data['target'] = strtolower( $data['target'] );
}
```

inc/class-statify-dashboard.php
```diff
@@ -385,4 +385,24 @@ private static function _select_data() {
...
+	/**
+	 * Parses a visit target and, if needed, modifies it for displaying.
+	 *
+	 * @param string $target URL target as displayed in widget.
+	 *
+	 * @return string
+	 */
+	public static function parse_target( $target ) {
+		// Modify for search strings.
+		if ( 0 === strpos( $target, '/?s=' ) ) {
```
stklcode: Are we sure the search parameter is always first, so there's no `...?foo=bar&s=searchstring`?
WP Core uses `isset( $this->query['s'] )` for that, so at this point it's position agnostic.

I'm fine with keeping it as simple and lightweight as possible here. But if that proves unreliable we might be better off extending this check or going the `wp_parse_url` / `wp_parse_args` way.

inc/class-statify-frontend.php
```diff
@@ -92,11 +92,23 @@ public static function track_visit( $is_snippet = false ) {
...
+		$data['target'] = str_replace( home_url( '/', 'relative' ), '/', $target );
+		// Maybe add trailing slash if that is no search query.
+		if ( 0 !== strpos( $data['target'], '/?s=' ) ) {
+			$data['target'] = user_trailingslashit( $data['target'] );
+		}
 
 		// Trim target url.
 		if ( $wp_rewrite->permalink_structure ) {
-			$data['target'] = wp_parse_url( $data['target'], PHP_URL_PATH );
+			if ( 0 !== strpos( $data['target'], '/?s=' ) ) {
```
stklcode: Same as the above comment on `Statify_Dashboard::parse_target`. Are we sure, the `s` parameter is always first?

At this point we are working on the actual request target, so maybe we could just rely on `is_search()` here which should already be populated at this point.

stklcode (changes requested):

inc/class-statify-dashboard.php
```diff
@@ -385,4 +385,24 @@ private static function _select_data() {
...
+			$target = preg_replace( '/^\/\?s=/', '', $target );
+			$target = urldecode( $target );
+			return sprintf( /* translators: s = search term */
+				__( 'Search query: “%s”', 'statify' ),
```
stklcode: Maybe just use `Search: %s` user input may get long anyway.

Zodiac1978:
Have you discussed about creating an own section for this instead of adding this data to the "views" section?

stklcode:
I didn't (maybe other parts of the team did). But a dedicated section seems like a good idea and might integrate well into the planned extended evaluation features. Should be removed from the pages view then or maybe just grouped together as "Search".

inc/class-statify-dashboard.php
florianbrinkmann: oh, good point, will change that

florianbrinkmann:
Didn’t think about adding an own section, but like the idea, then we also do not need the `Search query:` prefix

inc/class-statify-frontend.php
florianbrinkmann: Good idea, modified that

inc/class-statify-frontend.php
florianbrinkmann: Changed that and introduced a new method that checks a query string for a search query.

inc/class-statify-dashboard.php
florianbrinkmann: While working on that: we would also need to strip away any other query params except the search string, right (except the `p` query param, maybe)?

florianbrinkmann: Just tested it, no need for checking for the page query param, because if that exists, it overrules the search template, of course

inc/class-statify-dashboard.php
florianbrinkmann: Resolving that via the own search section

florianbrinkmann:
Changed it so that the search queries are displayd in its own list

Screenshot of the Statify widget with the new area for search queries

views/widget-front.php
```diff
@@ -34,11 +34,56 @@ class_exists( 'Statify' ) || exit;
...
+							</a>
+						</td>
+					</tr>
+=======
 		<p class="sub"><?php esc_html_e( 'Top targets', 'statify' ); ?></p>
 		<table>
 			<tbody>
 				<?php for ( $i = 0; $i < $limit; $i++ ) { ?>
 					<tr class="placeholder"><td colspan="2">&nbsp;</td></tr>
+>>>>>>> develop
```
2ndkauboy: There is another conflict in the latest commit.

florianbrinkmann: Thanks, fixed it

inc/class-statify-dashboard.php
```diff
@@ -301,8 +301,9 @@ public static function get_stats( $force_refresh = false ) {
 	/**
 	 * Get stats from DB
 	 *
-	 * @since    0.1.0
-	 * @version  1.4.0
+	 * @since   0.1.0
+	 * @since   1.4.0
+	 * @since   2.0.0  Add search queries to `$data` array.
```
2ndkauboy, line 306: Is this common to add a "changelog" like this to PHPDoc? And now we also have three `@since` and no `@version` anymore. :thinking:

florianbrinkmann: I guess we used the `@version` wrong, if I am right, it is ment to set versions for APIs and similar (https://docs.phpdoc.org/guide/references/phpdoc/tags/version.html#version), and you can use multiple `@since` (https://docs.phpdoc.org/guide/references/phpdoc/tags/since.html#since)

stklcode: Yep, `@since` is commonly used for changelog purposes, mostly changes to method signature or high-level behavior (at least helpful on public API that can be quite helpful to have this information on hand in the IDE)
`@version` is meant for semantic changes and the state of the functionality.

Multiple of either one don’t make much sense, because they yield barely any information, just “something was done here“.

Something like this would be more or less meaningful:
```
@since 1.0 introduced (if this is left empty, the natural assumption would be the same)
@since 1.1 return boolean result instead of void
@since 1.2 make parameter $foo optional with default 42

@version 1.1
```

I personally often omit the version tags, if there is no real need for that. The since-tags often help even in the same project to see why a method is used in a non-optimal way or similar. In such cases the hint “method was updated in 1.2“ and “we use it since 1.1“ immediately presents a valid reason, so we don’t have to guess what someone may have thought here.

Btw. that is also done in WP core and the changelogs are rendered in the reference articles, e.g.: https://developer.wordpress.org/reference/functions/wp_parse_url/#changelog

2ndkauboy: "Since" we only use it on some functions and we don't even have a real "public API", would we even mind keeping these `@since` version numbers up to date?

views/widget-front.php
```diff
@@ -44,6 +44,17 @@ class_exists( 'Statify' ) || exit;
 		</table>
 	</div>
 
+	<div class="table searches">
+		<p class="sub"><?php esc_html_e( 'Top searches', 'statify' ); ?></p>
+		<table>
+			<tbody>
+				<?php for ( $i = 0; $i < $limit; $i++ ) { ?>
+					<tr class="placeholder"><td colspan="2">&nbsp;</td></tr>
+				<?php } ?>
```
2ndkauboy, line 53: Do I get this right, that you first create a table with empty `<tr>` tags and then fill/replace them using JavaScript? Why not generate the `<table>` in JavaScript instead?

2ndkauboy: OK, you do just the same as with the other tables :see_no_evil:

inc/class-statify-dashboard.php
stklcode: Well, a public method is a public method and there are some recommendations out there for things like workarounds or custom triggers that call some of these. Some snippets given by ourselves in the support forums. Documenting API changes enables users of those to potentially migrate their logic without opening another support request, if changes are not self-explaining.

I agree, the majority of our code is designed for internal use and only exposed public for hooks etc.

But as with every project, that’s yet another part of documentation that might get out of sync over time. Though it’s probably one of the easier parts, as old annotations typically never become wrong, so we only have to append another one-liner, if necessary.

PR #247, opened 2023-03-19
Replaced bot detection with JayBizzle/Crawler-Detect

### LINKED ISSUE #217: Optimize bot detection
At the moment we use some string to detect crawler from the user agent string:

https://github.com/pluginkollektiv/statify/blob/667518428b30b0522367fb2c955d1913e1ef672f/inc/class-statify-frontend.php#L222-L236

We could add some more strings, like `seo`, `crawling` and `chrome-lighthouse`
 (borrowed from Koko Analytics):

https://github.com/ibericode/koko-analytics/blob/18716dc9156a83e72b2967cec6dee8ce9acfdbe9/assets/src/js/script.js#L53

Looking at the biggest 10 crawlers, I think we get almost all. But Alexa is missing with their `ia_archiver`.

Maybe something like `fetcher`and `scraper` too ...

`facebookexternalhit` could be another candidate.

Or we could take the big step and use a Third-Party-Library like https://github.com/JayBizzle/Crawler-Detect to detect crawlers.

Another one (in JSON) would be https://github.com/monperrus/crawler-user-agents/blob/master/crawler-user-agents.json

00Sleepy:
Implemented Composer Autoload for JayBizzle/Crawler-Detect and replaced bot detection in class-statify-frontend.php with CrawlerDetect function.

Solution for https://github.com/pluginkollektiv/statify/issues/217

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,9 +9,10 @@
     "statistics"
   ],
   "require": {
-    "php": "^5.2|^7|^8",
+    "php": "^5.3|^7|^8",
     "npm-asset/chartist": "^1.3.0",
-    "npm-asset/chartist-plugin-tooltips-updated": "^1.0.0"
+    "npm-asset/chartist-plugin-tooltips-updated": "^1.0.0",
+    "jaybizzle/crawler-detect": "^1.2"
   },
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -184,7 +184,7 @@ class Statify_Frontend extends Statify {
 		$user_agent = sanitize_text_field( isset( $_SERVER['HTTP_USER_AGENT'] ) ? wp_unslash( $_SERVER['HTTP_USER_AGENT'] ) : '' );
 		if ( is_null( $user_agent )
 			|| false === $user_agent
-			|| self::is_bot( $user_agent ) ) {
+			|| self::is_bot() ) {
 			return true;
 		}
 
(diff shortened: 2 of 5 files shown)


inc/class-statify-frontend.php
```diff
@@ -217,28 +217,10 @@ private static function _skip_tracking() {
...
-			'wget',
-			'monitor',
-		);
-
-		foreach ( $identifiers as $identifier ) {
-			if ( strpos( $user_agent, $identifier ) !== false ) {
-				return true;
-			}
-		}
+		$CrawlerDetect = new \Jaybizzle\CrawlerDetect\CrawlerDetect;
```
2ndkauboy: I'de rename this to `$crawler_detect` following the WordPress coding standards.

statify.php
```diff
@@ -53,6 +53,8 @@
 	)
 );
 
+/* Composer Autpload */
+require __DIR__ . '/vendor/autoload.php';
 
 /* Autoload */
 spl_autoload_register( 'statify_autoload' );
```
2ndkauboy: If we now use the Composer autoloader adding this feature, it might make sense to remove our own autoloader. But this could be done in another PR once this is merged.

inc/class-statify-frontend.php
2ndkauboy, line 220: $crawler_detect = new \Jaybizzle\CrawlerDetect\CrawlerDetect;

2ndkauboy:
```
$crawler_detect = new \Jaybizzle\CrawlerDetect\CrawlerDetect;
```

2ndkauboy (dismissed):

00Sleepy (commented):
Changed $CrawlerDetect  to $crawler_detect

2ndkauboy:
Looks good so far, can you please squash all commits? Alternatively we can squash it, when merging the PR.

2ndkauboy:
Now it looks good. @florianbrinkmann @krafit what do you think?

2ndkauboy (approved):

composer.json
```diff
@@ -23,6 +24,11 @@
     "phpunit/phpunit": "^5|^7|^9",
     "yoast/phpunit-polyfills": "^1.0"
   },
+  "autoload": {
+    "psr-4": {
+        "Jaybizzle\\CrawlerDetect\\": "src/"
+    }
+  },
```
2ndkauboy: I believe this is not necessary. Or was this documented in the package's README?

00Sleepy: Removed it an tested, works without outoload in the composer.json :-D

PR #251, opened 2023-03-19
WIP: Implement tracking of page title

### LINKED ISSUE #78: Show title instead of permalink

### LINKED ISSUE #196: Optimizing creation/config of SQL tables
There is a reported issue with creating the tables in the forums:
https://wordpress.org/support/topic/keine-anziege-von-daten/
Unfortunately the person is not tech savy and switched already to another plugin. No further information besides the (small) hoster name.

And I also have found an issue on a client website:
The table is set to "latin1_german1_ci", which could be the default on the server. I was wondering if we should set this to UTF-8 or not.

jujoko7CF:
User has a feature request to show the page title instead of just the target url. After some long discussions we decided to introduce a new statify meta table. In this table any additional tracking data can be saved. With this PR we save the wp_document_title in this meta table.

Fixes #78 , #196


inc/class-statify-frontend.php
```diff
@@ -87,24 +96,49 @@ public static function track_visit( $is_snippet = false ) {
...
+			if( array_key_exists( $meta_field['meta_key'], $tracking_meta ) ) {
+				$meta_value = $tracking_meta[ $meta_field['meta_key'] ];
+
+				$sanitize_function = isset( $meta_field['sanitize_callback'] ) && is_callable( $meta_field['sanitize_callback'] )
+					? $meta_field['sanitize_callback']
+					: 'sanitize_text_field';
+
+				$meta_value = call_user_func( $sanitize_function, $meta_value );
+
+				/* Init rows */
```
2ndkauboy: Single line comments should not use block comments.

inc/class-statify-schema.php
```diff
@@ -0,0 +1,119 @@
...
+		// Remove.
+		foreach( static::$tables as $table ) {
+			$wpdb->query( "DROP TABLE IF EXISTS `{$wpdb->$table}`" );
+		}
+
+		delete_option( 'statify_db_version' );
+	}
+}
+
+/* neue option: db Version 2.0.0 */
```
2ndkauboy: Remove this comment.

inc/class-statify-schema.php
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * @since 2.0.0
+ */
+class Statify_Schema {
+	/**
+	 * Database tables
+	 *
+	 * @var string[]
+	 */
+	public static $tables = array(
```
2ndkauboy, line 25: Maybe this should not be public and/or constant?

stklcode: Const array is PHP 5.6+ iirc (currently supported 5.3+)
But feel free to raise the requirements, if that make things easier here.

2ndkauboy: Constant arrays is not a thing at all :laughing: Like not before PHP 7+, so we just make it private.

stklcode:
Just a reminder that this change conflicts with the planned aggregation (#234) or at least requires significant changes on at least one side.

2ndkauboy:
@stklcode it is true that this might conflict with the data aggregation. But not necessarily. We haved discussed the data aggregation being in conflict with multiple other issues and have to yet decided if, and how the data aggregation would be done. We could still have aggregated data while still collecting more information. We "only" need to decide "at aggregation" how to handle this aggregated data. Would we aggregate per unique touple, use just one (the latest) additional data set? This should then be discussed in the data aggregation ticket.

stklcode:
I didn’t say that aggregation is impossible with additional metadata, just not in the currently proposed way, i.e. we cannot simply merge both without additional work.

The title typically has 1-to-1 relation with the target URL at at specific point in time, so this case is more or less trivial to solve. With more attributes combinations grow and the benefits of tuple aggregation is thus reduced, so multi-layer aggregation over each attribute may be worth thinking of. But this would be incompatible with the proposed data structure.

But all this might be overthought at this time, we should not make it more complex than necessary for the first shot.

Larzans:
Is this still in the pipeline?
It would be great to have a hook in statify that we can call with extra data that then gets stored in the statifymeta table for easier extension.
The evaluation could then be done with an extension so we wouldn't have to touch the statify code.

PR #255, opened 2023-03-20
Improve the field explanation for "Disallowed Comment Keys" (#245)

### LINKED ISSUE #245: Improve field explanaition
I didn't really knew what the checkbox "Disallowed referrers" would do. It is not really explained in a way, that it's clear what it does, especially in the current German translation. It referrs to the "Disallowed Comment Keys" field in the "Discussion" settings. Maybe we can link the term "Disallowed referrers" to this settings field and explain it better?

2ndkauboy:
Closes #245


florianbrinkmann (approved):
LGTM

PR #257, opened 2023-03-27
integrate extended evaluation

stklcode:
Integrate dashboards from _Extended Evaluation for Statify_ by @patrickrobrecht 
(https://github.com/patrickrobrecht/extended-evaluation-for-statify) 

We include the queries without any semantic change and introduce 3 new admin pages.
Rendering of charts and tables is done client-side, as we now do for the small dashboard widget, too and all data is retrieved via WP API.

**This is Work in Progress**

* Backend/API:
  * [x] Yearly/monthly values
  * [x] Daily values
  * [x] Filter by post/page
  * [x] Top targets
  * [x] Top referrers
* UI:
  * [x] Yearly/monthly overview
  * [x] Per-year tabs with monthly/daily values
  * [x] Filter by post/page
  * [x] Top targets
  * [x] Top referrers
  * [x] CSV export

**SNAPSHOT Builds:**

* ~statify.2.0.0-snap257-20250824.zip~
* ~statify.2.0.0-snap257-20250826.zip~
* ~[statify.2.0.0-snap257-20250828.zip](https://github.com/user-attachments/files/22019734/statify.2.0.0-snap257-20250828.zip)~
* [statify.2.0.0-snap257-20260315.zip](https://github.com/user-attachments/files/26005508/statify.2.0.0-snap257-20260315.zip)

(use at own risk)

----
statfy257_001
statfy257_002

statfy257_003

### DIFF
--- a/inc/class-statify-api.php
+++ b/inc/class-statify-api.php
@@ -22,10 +22,11 @@ class Statify_Api extends Statify {
 	 *
 	 * @var    string
 	 */
-	const REST_NAMESPACE   = 'statify/v1';
+	const REST_NAMESPACE = 'statify/v1';
 	const REST_ROUTE_TRACK = 'track';
 	const REST_ROUTE_STATS = 'stats';
 	const REST_ROUTE_RESET = 'reset';
+	const REST_ROUTE_STATS_EXTENDED = 'stats/extended';
 
 	/**
 	 * Initialize REST API routes.
@@ -65,6 +66,16 @@ class Statify_Api extends Statify {
 				'permission_callback' => function () {
 					return current_user_can( 'manage_options' );
 				},
+			)
+		);
+
+		register_rest_route(
+			self::REST_NAMESPACE,
+			self::REST_ROUTE_STATS_EXTENDED,
+			array(
+				'methods'             => WP_REST_Server::READABLE,
+				'callback'            => array( __CLASS__, 'get_extended' ),
+				'permission_callback' => array( __CLASS__, 'user_can_see_stats' ),
 			)
 		);
 
@@ -116,7 +127,7 @@ class Statify_Api extends Statify {
 			if ( null !== $referrer ) {
 				$referrer = filter_var( $referrer, FILTER_SANITIZE_URL );
 			}
-			$target   = $request->get_param( 'target' );
+			$target = $request->get_param( 'target' );
 			if ( null !== $target ) {
 				$target = filter_var( $target, FILTER_SANITIZE_URL );
 			}
@@ -137,7 +148,7 @@ class Statify_Api extends Statify {
 	public static function get_stats( WP_REST_Request $request ): WP_REST_Response {
 		$refresh = '1' === $request->get_param( 'refresh' );
 
-		$stats  = Statify_Dashboard::get_stats( $refresh );
+		$stats = Statify_Dashboard::get_stats( $refresh );
 
 		$visits          = $stats['visits'];
 		$stats['visits'] = array();
@@ -192,5 +203,54 @@ class Statify_Api extends Statify {
 			),
 			200
 		);
+	}
+
+	/**
+	 * Get extended statistics.
+	 *
+	 * @param WP_REST_Request $request The request.
+	 *
+	 * @return WP_REST_Response The response.
+	 *
+	 * @since 2.0.0
+	 */
+	public static function get_extended( WP_REST_Request $request ): WP_REST_Response {
+		$scope = $request->get_param( 'scope' );
+		$post  = $request->get_param( 'post' );
+
+		$status = 200;
+		if ( 'year' === $scope ) {
+			$stats = Statify_Evaluation::get_views_for_all_years( $post );
+		} elseif ( 'month' === $scope ) {
+			$visits  = Statify_Evaluation::get_views_for_all_months( $post );
+			$stats   = array( 'visits' => array() );
+			$last_ym = 0;
+			foreach ( $visits as $ym => $v ) {
+				$ym         = explode( '-', $ym );
+				$year       = intval( $ym[0] );
+				$month      = intval( $ym[1] );
+				$year_month = $year * 12 + $month;
+				for ( $ym = $last_ym + 1; $last_ym > 0 && $ym < $year_month; $ym++ ) {
+					// Fill gaps.
+					$y = intval( $ym / 12 );
+					if ( ! isset( $stats['visits'][ $y ] ) ) {
+						$stats['visits'][ $y ] = array();
+					}
+					$stats['visits'][ $y ][ $ym % 12 ] = 0;
+				}
+				if ( ! isset( $stats['visits'][ $year ] ) ) {
+					$stats['visits'][ $year ] = array();
+				}
+				$stats['visits'][ $year ][ $month ] = $v;
+				$last_ym                            = $year_month;
+			}
+		} elseif ( 'day' === $scope ) {
+			$stats = Statify_Evaluation::get_views_for_all_days( $post );
+		} else {
+			$stats  = array( 'error' => 'invalid scope (allowed: year, month, day)' );
+			$status = 400;
+		}
+
+		return new WP_REST_Response( $stats, $status );
 	}
 }
--- a/inc/class-statify-api.php
+++ b/inc/class-statify-api.php
@@ -215,42 +215,83 @@ class Statify_Api extends Statify {
 	 * @since 2.0.0
 	 */
 	public static function get_extended( WP_REST_Request $request ): WP_REST_Response {
+		// Verify scope.
 		$scope = $request->get_param( 'scope' );
-		$post  = $request->get_param( 'post' );
-
-		$status = 200;
-		if ( 'year' === $scope ) {
-			$stats = Statify_Evaluation::get_views_for_all_years( $post );
-		} elseif ( 'month' === $scope ) {
-			$visits  = Statify_Evaluation::get_views_for_all_months( $post );
-			$stats   = array( 'visits' => array() );
-			$last_ym = 0;
-			foreach ( $visits as $ym => $v ) {
-				$ym         = explode( '-', $ym );
-				$year       = intval( $ym[0] );
-				$month      = intval( $ym[1] );
-				$year_month = $year * 12 + $month;
-				for ( $ym = $last_ym + 1; $last_ym > 0 && $ym < $year_month; $ym++ ) {
-					// Fill gaps.
-					$y = intval( $ym / 12 );
-					if ( ! isset( $stats['visits'][ $y ] ) ) {
-						$stats['visits'][ $y ] = array();
-					}
-					$stats['visits'][ $y ][ $ym % 12 ] = 0;
-				}
-				if ( ! isset( $stats['visits'][ $year ] ) ) {
-					$stats['visits'][ $year ] = array();
-				}
-				$stats['visits'][ $year ][ $month ] = $v;
-				$last_ym                            = $year_month;
-			}
-		} elseif ( 'day' === $scope ) {
-			$stats = Statify_Evaluation::get_views_for_all_days( $post );
-		} else {
-			$stats  = array( 'error' => 'invalid scope (allowed: year, month, day)' );
-			$status = 400;
+		if ( ! in_array( $scope, array( 'year', 'month', 'day' ) ) ) {
+			return new WP_REST_Response(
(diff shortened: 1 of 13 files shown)


inc/class-statify-api.php
```diff
@@ -151,4 +195,281 @@ public static function get_stats( WP_REST_Request $request ): WP_REST_Response {
...
+			$yr = intval( $yr );
+			if ( $yr <= 0 ) {
+				return new WP_REST_Response(
+					array( 'error' => 'invalid year' ),
+					400
+				);
+			}
+		} else {
+			$yr = 0;
+		}
```
florianbrinkmann:
```suggestion
		$yr = $request->get_param( 'year' ) ?? 0;
		if ( ! empty( $yr ) ) {
			$yr = intval( $yr );
			if ( $yr <= 0 ) {
				return new WP_REST_Response(
					array( 'error' => 'invalid year' ),
					400
				);
			}
		}
```

inc/class-statify-evaluation.php
```diff
@@ -0,0 +1,640 @@
...
+	/**
+	 * Returns the average daily views in the given month.
+	 *
+	 * @param array $views_for_all_months an array with the monthly views.
+	 * @param int   $year the year.
+	 * @param int   $month the month.
+	 *
+	 * @return int the average daily views in the given month.
+	 */
+	public static function get_average_daily_views_of_month( array $views_for_all_months, int $year, int $month ): int {
```
florianbrinkmann: I don’t find where those `get_average_` methods are used (except in the tests). Are those obsolete or am I overseeing something? 🫣

inc/class-statify-evaluation.php
```diff
@@ -0,0 +1,640 @@
...
+	/**
+	 * Returns the views for one year. If the date does not exist (e.g. 30th February),
+	 * this method returns -1.
+	 *
+	 * @param array $views_for_all_years an array with the yearly views.
+	 * @param int   $year the year.
+	 *
+	 * @return int the views of the given year.
+	 */
+	public static function get_yearly_views( array $views_for_all_years, int $year ): int {
```
florianbrinkmann: Can’t find usage of this function either

inc/class-statify.php
```diff
@@ -225,6 +234,102 @@ public static function user_can_see_stats(): bool {
...
+
+	/**
+	 * Print CSS
+	 *
+	 * @since 0.1.0
+	 * @since 2.0.0 Moved to Statify class
+	 */
+	public static function add_style(): void {
+
+		// Register CSS.
```
florianbrinkmann:
```suggestion
		// Register CSS.
```

js/dashboard.js
```diff
@@ -19,55 +120,184 @@
...
+					renderTotals(tables.totals, data.totals);
 				}
 			})
 			.catch(() => {
 				// Failed to load.
-				chartElem.innerHTML =
+				charts.dashboard.innerHTML =
 					'<p>' +
 					wp.i18n.__('Error loading data.', 'statify') +
 					'</p>';
```
florianbrinkmann: Should be possible to write it shorter like that: 

```suggestion
				charts.dashboard.innerHTML = `<p>${wp.i18n.__('Error loading data.', 'statify')}</p>`;
```

js/dashboard.js
```diff
@@ -159,18 +390,94 @@
...
+			root,
+			{
+				labels,
+				series: [values],
+			},
+			{
+				low: 0,
+				showArea: true,
+				// fullWidth: true,
+				// width: fullWidth ? undefined : 5 * labels.length,
```
florianbrinkmann:
```suggestion
```

js/dashboard.js
```diff
@@ -159,18 +390,94 @@
...
 	 */
 	function renderTopList(table, data) {
-		// Get pre-existing rows.
-		const rows = table.querySelectorAll('tr');
-
-		// Update or append rows.
-		data.forEach((r, idx) => {
+		const rows = data.map((r) => {
 			const row = document.createElement('TR');
 			row.innerHTML =
```
florianbrinkmann, line 489: Could use interpolation, too.

```
row.innerHTML = `<td class="b">${r.count}</td><td class="t"><a href="${r.url}" target="_blank" rel="noopener noreferrer">${(r.host || r.url)}</td>`;
```

js/dashboard.js
```diff
@@ -201,22 +501,17 @@
 	 * @param {{alltime: number, since: string, today: number}} data  Totals data.
 	 */
 	function renderTotals(table, data) {
-		const rows = table.querySelectorAll('tr');
-		let row = document.createElement('TR');
-		row.innerHTML =
+		const rowToday = document.createElement('TR');
+		rowToday.innerHTML =
```
florianbrinkmann, line 506: Same as above:

```
rowToday.innerHTML = `<td class="b">${data.today}</td><td class="t">${wp.i18n.__('today', 'statify')}</td>`;
```

js/dashboard.js
```diff
@@ -201,22 +501,17 @@
...
-		if (rows.length > 0) {
-			table.replaceChild(row, rows[0]);
-		} else {
-			table.appendChild(row);
-		}
-		row = document.createElement('TR');
-		row.innerHTML =
+
+		const rowAll = document.createElement('TR');
+		rowAll.innerHTML =
```
florianbrinkmann, line 511:
```
rowAll.innerHTML = `<td class="b">${data.alltime}</td><td class="t">${wp.i18n.sprintf(
				/* translators: %s: Date. */
				wp.i18n.__('since %s', 'statify'),
				data.since
			)}</td>`;
```

js/dashboard.js
```diff
@@ -252,4 +790,147 @@
...
+
+	/**
+	 * Add a CSV export button to a table.
+	 *
+	 * @param {HTMLTableElement} table Table to process
+	 */
+	function addExportButton(table) {
+		const exportBtn = document.createElement('a');
+		exportBtn.classList.add('button');
+		exportBtn.role = 'button';
```
florianbrinkmann: As the `href` attribute is used here to trigger the download, in my opinion it’s a link and not a button, so shouldn’t get the `button` role (and if it would be a button semantically, should use the `button` element).

The problem with setting the `href` on click is, that a link is not focusable via e.g. keyboard without an `href`. Would it be possible to populate the `href` on load?

js/dashboard.js
```diff
@@ -252,4 +790,147 @@
...
+	 * @param {string}                                             path  Relative API path.
+	 * @param {string|Record<string, string>|URLSearchParams|null} param Parameters (optional).
+	 * @return {Promise<any>} Response promise.
+	 */
+	function fetchData(path, param = undefined) {
+		return wp.apiFetch({
+			path:
+				'/statify/v1/' +
+				path +
+				(param ? '?' + new URLSearchParams(param) : ''),
```
florianbrinkmann:
```suggestion
			path: `/statify/v1/${path}${(param ? '?' + new URLSearchParams(param) : '')}`,
```

views/view-content.php
```diff
@@ -0,0 +1,171 @@
...
+		</a>
+		<?php foreach ( $post_types as $pt ) : ?>
+		<a href="<?php echo esc_url( admin_url( 'admin.php?page=statify_content&type=' . $pt ) ); ?>"
+		   class="nav-tab<?php echo ( $selected_type === $pt ) ? ' nav-tab-active' : ''; ?>">
+			<?php echo esc_html( get_post_type_object( $pt )->labels->name ); ?>
+		</a>
+		<?php endforeach; ?>
+	</nav>
+
+	<h2></h2>
```
florianbrinkmann:
```suggestion
```

views/view-content.php
```diff
@@ -0,0 +1,171 @@
...
+		<input id="statify-content-datestart" name="start" type="date" value="<?php echo esc_attr( $date_start ); ?>">
+		<label for="statify-content-dateend">End date</label>
+		<input id="statify-content-dateend" name="end" type="date" value="<?php echo esc_attr( $date_end ); ?>">
+		<button type="submit" class="button-secondary"><?php esc_html_e( 'Select date period', 'statify' ); ?></button>
+	</form>
+
+	<?php if ( 'popular' === $selected_type ) : ?>
+
+	<section>
+		<h3><?php esc_html_e( 'Most Popular Content', 'statify' ); ?></h3>
```
florianbrinkmann: Because there’s no `h2` before, this needs to be one

```suggestion
		<h2><?php esc_html_e( 'Most Popular Content', 'statify' ); ?></h2>
```

views/view-content.php
```diff
@@ -0,0 +1,171 @@
...
+		<div class="statify-chart-container">
+			<div id="statify_chart_content" class="statify-chart">
+				<span class="spinner is-active" title="<?php esc_html_e( 'loading', 'statify' ); ?>"></span>
+			</div>
+		</div>
+	</section>
+	<?php endif; ?>
+
+	<section>
+		<h3><?php echo esc_html( $section_title ); ?></h3>
```
florianbrinkmann:
```suggestion
		<h2><?php echo esc_html( $section_title ); ?></h2>
```

views/view-referrers.php
```diff
@@ -0,0 +1,138 @@
...
+	<section>
+		<div class="statify-chart-container">
+			<div id="statify_chart_referrer" class="statify-chart">
+				<span class="spinner is-active" title="<?php esc_html_e( 'loading', 'statify' ); ?>"></span>
+			</div>
+		</div>
+	</section>
+
+	<section>
+		<h3><?php echo esc_html( $section_title ); ?></h3>
```
florianbrinkmann:
```suggestion
		<h2><?php echo esc_html( $section_title ); ?></h2>
```

views/view-referrers.php
```diff
@@ -0,0 +1,138 @@
...
+		/* translators: %s is replaced by a post title or "all posts" */
+		__( 'Referrers from other websites for %s', 'statify' ),
+		$selected_post_title
+	);
+}
+?>
+<div class="wrap">
+	<h1><?php esc_html_e( 'Statify', 'statify' ); ?> - <?php esc_html_e( 'Referrers from other websites', 'statify' ); ?></h1>
+
+	<br>
```
florianbrinkmann: If it’s not too cumbersome, maybe replace this and the other `<br>` with CSS?

inc/class-statify-api.php
stklcode: :+1: Missed that one, original method was written before PHP 7

js/dashboard.js
stklcode: Updated in all 4 places (3 new + 1 pre-existing). Same for some `'<td>' + ... + '</td>'` blocks.

views/view-referrers.php
stklcode: Indeed, styling with breaks is not very elegant. Added `margin: 1em 0` CSS to the `<form id="statify-dashboard-controls">` element to deal with spacing here.

js/dashboard.js
stklcode: > As the `href` attribute is used here to trigger the download, in my opinion it’s a link and not a button, so shouldn’t get the button role (and if it would be a button semantically, should use the `button` element).

Yes, good point. We should drop the role here.

> The problem with setting the `href` on click is, that a link is not focusable via e.g. keyboard without an `href`. Would it be possible to populate the `href` on load?

Yes, the link/button is added at the end of the render method, so data is available at this point.

I see valid arguments for either using empty or `#` target initially, just to make it focusable, and pre-generating the CSV. It's not really expensive, but I guess not the most popular feature, so typically not required, that's why I made it lazy initially.

inc/class-statify-evaluation.php
stklcode: These methods are taken from the _Extended Evaluation_ plugin and originally used to fill the average columns in PHP-rendered tables.

We now calculate average and sum in the frontend JS and should clean up the orphaned backend logic.

stklcode: Done. Removed 8 methods that were only called from tests or internally.

* `get_days( $month, $year )`
* `get_daily_views( $views_for_all_days, $year, $month, $day )`
* `get_monthly_views( $views_for_all_months, $year, $month )`
* `get_average_daily_views_of_month( $views_for_all_months, $year, $month )`
* `get_daily_views_of_month( $views_for_all_days, $year, $month )`
* `is_current_month( $year, $month )`
* `get_yearly_views( $views_for_all_years, $year )`
* `get_views_of_post( $url, $start, $end )`

PR #260, opened 2023-03-31
copy third-party libs instead of using Composer autoloader (#256)

### LINKED ISSUE #256: Composer dependencies and autoloader missing in deployment
**Describe the bug**

Regression from #247

As part of this change we introduced an external library for crawler detection and use the Composer autoloader mechanism for that.

However, the _vendor_ directory is currently excluded from deployment and thus missing in the final package.

**To Reproduce**
1. Build the project as it's done in the CI pipeline
2. Install the plugin
3. Visit any page
4. See PHP errors due to missing _.../vendor/autoload.php_

**Expected behavior**
Page loads as before without errors.

- any

**Additional context**
Plugin from current _develop_ branch (starting from c5c21badd784861be6860b9e42659ddd884eb1b4)

stklcode:
fixes #256 

The  autoloader mechanism is quite heavy for a single dependency and requires _composer_ and _vendor_ dir to be part of the final product.

We yet have generated/minified resources in our project, so this is the minimalistic approach to copy the _CrawlerDetect_ lib, too and extend the custom autoloader mechanism.

----

This is not a pretty solution, but the project is not really _Composer_ compatible. (no PSR-4 namespaces, requirement for minified files that are generated from dev-dependencies, etc.).

This way we ca postpone the decision to heavily rework the project structures while having a runnable artifact again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ coverage
 css/*.min.css
 js/*.min.js
 js/*.min.js.map
+lib
 phpunit.*.xml
 vendor/
 node_modules/
--- a/composer.json
+++ b/composer.json
@@ -83,6 +83,7 @@
   },
   "extra": {
     "copy-file": {
+      "vendor/jaybizzle/crawler-detect/src/": "lib/Jaybizzle/CrawlerDetect/",
       "vendor/npm-asset/chartist/dist/index.css": "css/chartist.min.css",
       "vendor/npm-asset/chartist/dist/index.umd.js": "js/chartist.min.js",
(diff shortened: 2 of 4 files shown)


florianbrinkmann:
The only thing missing here would be to add the CrawlerDetect file(s) to the `lib` folder, right?

stklcode:
You're right, seems the line got lost at some point. Re-added the missing directory to the `copy-file` list in _composer.json_, so it's populated after `composer install`

florianbrinkmann (approved):
LGTM!

PR #261, opened 2023-03-31
migrate to PSR-4 namespace and use Composer in production

stklcode:
This PR is **WORK IN PROGRESS** and not yet ready to be merged.

----

This proposal does change the project structure and class names.

First we introduce a `Pluginkollektiv\Statify\` namespace and move the classes into it stripping their prefixes, i.e. `Statify_Frontend` will become `Pluginkollektiv\Statify\Frontend`.
Then we introduce a PSR-4 autoloader with _Composer_ and replace our custom autoloading function.

ToDo:
* Find a different way to minify our CS/JS resources, because this currently requires dev-dependencies, so the mechanism is not yet suitable for production.
  * Might be a pre-commit hook that generates them, so they are kept in the source tree while still be automatically generated.
* Point third-party CSS/JS to the vendor directory.
* Discuss, whether we want to go that way (probably do this first place)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -88,6 +88,11 @@
       "vendor/npm-asset/chartist-plugin-tooltips-updated/dist/chartist-plugin-tooltip.min.js": "js/"
     }
   },
+  "autoload": {
+    "psr-4": {
+      "Statify\\": "inc/"
+    }
+  },
   "config": {
     "allow-plugins": {
       "dealerdirect/phpcodesniffer-composer-installer": true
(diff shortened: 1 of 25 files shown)


chesio:
A small suggestion: I would stick to `Vendor\Product` scheme for namespaces as this scheme is well-established in PHP world and use `Pluginkollektiv` as vendor name, so the new namespace would be `Pluginkollektiv\Statify`. This lessens the risk of namespace collision to practically zero as Pluginkollektiv is also well-established name in PHP world... :-)

stklcode:
Good point.

florianbrinkmann:
I would vote to go with the minimal approach from #260 for now

PR #264, opened 2023-06-08
fix text alignment in dashboard widget (#263)

### LINKED ISSUE #263: Period before and not after message
Bildschirmfoto 2023-06-08 um 15 25 55

This is because of a CSS issue:

`#statify_chart {
  color: #a7aaad;
  direction: rtl;
  text-align: center;
}`

This is wrong `direction: rtl` and is causing the issue.

Thanks to @Drivingralle for reporting it to me.

More context:
https://github.com/pluginkollektiv/statify/pull/101
https://github.com/pluginkollektiv/statify/commit/9c9fc3d29dd55ebc9f6314a0d1cf543da6c6b2ab

stklcode:
We use RTL direction for the chart container in order see the rightmost data when the chart has a horizontal scrollbar. This however results in wrong text layout if there is no data or an error occurred.

Revert the text direction for all child elements and remove the custom flex overrides for the chart labels.

resolves #263 

----
The scrollbar is right by default:
statify-264-01

The labels are right-aligned:
statify-264-02

Inner text is displayed in default direction:
statify-264-03

And on sites with RTL languages, too:
statify-264-04

### DIFF
--- a/css/dashboard.css
+++ b/css/dashboard.css
@@ -3,7 +3,6 @@
 #statify_chart {
 	color: #a7aaad;
 	flex: 0 0 100%;
-	direction: rtl;
 	height: 140px;
 	margin: 1em -0.5em;
 	overflow-x: auto;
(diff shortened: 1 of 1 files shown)


Zodiac1978 (approved):
LGTM!

PR #271, opened 2023-09-22
Also create the `stats` API endpoint for no-JS tracking (#270)

### LINKED ISSUE #270: Always create `stats` API endpoint
Currently we just create the API endpoints if JS tracking is active, but the dashboard uses the `stats` endpoint so that needs to exist also for no-JS tracking.

florianbrinkmann:
Fixes #270


stklcode (approved):
LGTM, thanks for the fix.

PR #272, opened 2023-09-24
update dashboard JS to ES2015 with recommended WPCS style

stklcode:
Preserve ES5 for the tracking snippet for maximum compatibiltiy. The dashboard requires some modern logic for the Bakcbone API client already so let's migrate our script to ES2015 as well.

We also split the large update routine into multiple methods in preparation for additional visualizations.

This change finally drops IE11 support for the admin dashboard.


florianbrinkmann (approved):
LGTM

PR #275, opened 2023-11-05
remove XML-RPC integration

stklcode:
The XML-RPC feature is unmaintained and there is no known use case for the `statify.getStats` endpoint. With statistics data now available as a REST API endpoint we have a modern alternative in place. Let's remove the feature and make room for new stuff.


pfefferle (approved):
Makes sense!

PR #276, opened 2023-11-09
update WPCS to v3.0 with minor code style tweaks

stklcode:
* update WPCS definitions to v3.0
* use `printf()` instead of `echo sprintf()`
* use `__DIR__` instead of `dirname( __FILE__ )`
* remove unused parameter and inconsistend return value from `track_visit()`

### DIFF
--- a/inc/class-statify-frontend.php
+++ b/inc/class-statify-frontend.php
@@ -21,17 +21,15 @@ class Statify_Frontend extends Statify {
 	/**
 	 * Track the page view
 	 *
-	 * @since    0.1.0
-	 * @since    1.7.0 $is_snippet parameter added.
-	 * @version  1.7.0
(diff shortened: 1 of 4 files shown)


stklcode:
Apparently this was already done in b2abe7a800f3d64b10c17a88699b475325c9dd54 and the only remaining change is superseded by #286. Closing this PR.

lloc:
> * use `printf()` instead of `echo sprintf()`

Is there an advantage to using `echo sprint()` over `printf()`?

stklcode:
No, there is no benefit in generating a formatted string and echoing the output immediately.

There is a marginal difference from the outside:
* `echo sprintf(...)` has no return value
* `printf(...)` returns the output length (we don't use it)

PR #279, opened 2023-11-24
Fix user role error on saving settings

### LINKED ISSUE #278: User role error on save settings
**Describe the bug**
Saving the settings is currently giving an error because it's pushing a key to an unexistent user roles array.

**To Reproduce**
1. Go to Settings
2. Select whatever user roles you want
3. Save

**Expected behavior**
Reloading to the current Settings page, with the user roles as set in the previous save.

- Develpment branch

pedro-mendonca:
This PR adds an empty user role array before pushing any new user role.

Fixes #278


stklcode:
Thanks! ... and sorry for hijacking the PR :wink: 

Changes are good, just took the chance to add a unit test for the affected method (that fails without the change due to type errors)

stklcode (approved):

Issue #253, opened 2023-03-20
Update required WordPress and PHP versions for 2.0


stklcode:
We have to raise the required WordPress and PHP versions according to the development state for the next release.
The final decision should be something reasonable, likely WP 5.x and PHP 5.6 or maybe 7.x. Can be postponed to the last responsible moment, just adding this issue to the milestone here, so we won’t forget about it.

Current requirements (up to 1.8.4):
WordPress 4.7
PHP 5.2

Incomplete list of merged, but unreleased changes:
#247 - add a library requiring PHP 5.3
#252 - introduces hooks that require WP 5.1

florianbrinkmann:
I would be also fine with 5.3, but 5.6 or 7.x sounds better of course :)

@krafit @pfefferle

Zodiac1978:
WP 5.1 is the last version which works with PHP <5.6. 

WP 5.2 requires PHP 5.6.

WP 4.1 is the first version which supports PHP 5.6.

So we could require PHP 5.3 and WP 5.1 without any problem.

See: https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/

stklcode:
PHP 7.2 & WP 5.1 it is, see #287

PR #280, opened 2023-11-24
[I18n] Add JavaScript i18n, typo and brand names

pedro-mendonca:
This PR includes the following:

- [x] Add JavaScript i18n to proper deal with JavaScript strings, including plurals, variables, etc
- [x] Fix ~statify~ Statify brand name
- [x] Typo

### DIFF
--- a/inc/class-statify-dashboard.php
+++ b/inc/class-statify-dashboard.php
@@ -131,21 +131,8 @@ class Statify_Dashboard extends Statify {
 			true
 		);
 
-		// Localize strings.
-		wp_localize_script(
-			'statify_chart_js',
-			'statifyDashboard',
-			array(
-				'i18n'  => array(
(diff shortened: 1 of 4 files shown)


stklcode:
Thanks Pedro!

I think we should add `wp-i18n` to the JS dependencies explicitly. On the dashboard it should be present anyway, so it works without, but relying on that is probably not best practice, is it?`

https://github.com/pluginkollektiv/statify/blob/83a29f1ab18e3df5b9a964b417e234431e2b83ff/inc/class-statify-dashboard.php#L126-L132


ping #253 for another WP 5.0+ argument

pedro-mendonca:
@stklcode I've added the `wp-i18n` core script to the dependencies as you've requested.
Thanks for the review :)

stklcode (approved):
LGTM. Thanks @pedro-mendonca for taking care of that!

PR #252, opened 2023-03-19
Replace some deprecated hooks with their new recommended ones

jujoko7CF:
Replace `wpmu_new_blog` with `wp_initialize_site`.
Replace `delete_blog` with `wp_uninitialize_site`.

PR #287, opened 2024-03-16
update requirements to PHP 7.2 and WP 5.1

stklcode:
As discussed in #253 we already have some pending changes that require more recent PHP and WordPress versions.

Current requirements (up to 1.8.x):
* WordPress 4.7
* PHP 5.2

Minimum without reverting stuff would be PHP 5.3 (#247) and WP 5.1 (#252).

Maintaining and testing old stuff is time consuming and only very few sites actually have such requirements. With this change, we jump forward to:
* WordPress 5.1
* PHP 7.2

WP 5.1 does not play nice with PHPUnit 7/8 and neither with Polyfills, so let's just use 5.2 as minimum version in CI tests and rely on the static compatibility checks instead of investing time and effort in testing one more version.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@
 * Tags:              analytics, dashboard, pageviews, privacy, statistics, stats, visits, web stats, widget
 * Requires at least: 4.7
 * Tested up to:      6.3
-* Requires PHP:      5.3
+* Requires PHP:      7.2
 * Stable tag:        1.8.4
 * License:           GPLv3 or later
 * License URI:       https://www.gnu.org/licenses/gpl-3.0.html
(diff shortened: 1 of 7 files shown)


readme.txt
```diff
@@ -2,9 +2,9 @@
 * Contributors:      pluginkollektiv
 * Donate link:       https://www.paypal.com/cgi-bin/webscr?cmd=_donations&business=TD4AMD2D8EMZW
 * Tags:              analytics, dashboard, pageviews, privacy, statistics, stats, visits, web stats, widget
-* Requires at least: 4.7
-* Tested up to:      6.3
-* Requires PHP:      5.3
+* Requires at least: 5.1
+* Tested up to:      6.4
```
florianbrinkmann: Could we already raise that to 6.5?

florianbrinkmann (approved):

readme.txt
stklcode: Sure, why not. Works fine on a local 6.5 RC test site and even if there's something wrong with it, we should be able to fix this before release.

PR #289, opened 2024-03-24
introduce blueprint.json for plugin preview

stklcode:
Add a _blueprint.json_ to run WordPress in Playground with _Statify_ installed.

Ref.: https://developer.wordpress.org/plugins/wordpress-org/previews-and-blueprints/

### DIFF
--- a/.wordpress-org/blueprints/blueprint.json
+++ b/.wordpress-org/blueprints/blueprint.json
@@ -1,0 +1,17 @@
+{
+  "$schema": "https://playground.wordpress.net/blueprint-schema.json",
+  "landingPage": "/wp-admin/",
+  "steps": [
+    {
+      "step": "login",
+      "username": "admin",
+      "password": "password"
+    },
+    {
+      "step": "defineWpConfigConsts",
+      "consts": {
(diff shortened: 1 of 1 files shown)


Zodiac1978:
If I am correct, it is not necessary to install the plugin on WordPress.org. This is only needed, if you are coming from the outside. On wp.org this happens automatically.

See this working example from ASB:
https://github.com/pluginkollektiv/antispam-bee/blob/master/.wordpress-org/blueprints/blueprint.json

stklcode:
Hm I see it's working for ASB, so I assume it will do the job for Statify as well.

I tested the blueprint locally with `wp-now`, this seems to require explicit plugin setup to do anything useful. Seems implementations have some differences.

I just pushed another blueprint for _Statify Filter_ (https://wordpress.org/plugins/statify-blacklist/) with installation steps. Think it is necessary there, because the plugin has a dependency to _Statify_ and WP < 6.5 does not support dependencies out of the box. (and I have no idea if the playground supports it in 6.5)

Zodiac1978 (approved):
LGTM!

PR #295, opened 2024-10-30
Adjust parameters for compatibility with multisite hooks

zhyian:
Hello!

This PR updates the multisite related functionality to use the `WP_Site` object instead of a site ID parameter, providing compatibility with the [wp_initialize_site](https://developer.wordpress.org/reference/hooks/wp_initialize_site/) and [wp_uninitialize_site](https://developer.wordpress.org/reference/hooks/wp_uninitialize_site/) hooks.

**Current behavior:** 
When the plugin is network-activated, creating a new site in the network causes a fatal error:

Screenshot 2024-10-30 at 15 25 02

### DIFF
--- a/inc/class-statify-install.php
+++ b/inc/class-statify-install.php
@@ -51,11 +51,11 @@ class Statify_Install {
 	 *
 	 * @since 1.4.4
 	 *
-	 * @param int $site_id Site ID.
+	 * @param WP_Site $new_site Site object.
 	 */
-	public static function init_site( int $site_id ): void {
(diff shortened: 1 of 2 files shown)


stklcode:
Fixes a regression from #252 (switch from [`wpmu_new_blog`](https://developer.wordpress.org/reference/hooks/wpmu_new_blog/) which expected `int` site ID)

Thanks @zhyian for taking care of that!

stklcode (approved):

PR #308, opened 2025-12-21
[stable] declare compatibility with WP 6.9 and fix CI for stable releases

stklcode:
Just a simple readme update, but unfortunately some additional fixes are required for CI. _Asset Packagist_ fails to provide _Chartist_ 0.x dependencies, PHP 5.6 fails with current test framework, ...

Apply some tweaks, to the pipelines actually run. We disable PHP 5.6 CI tests for now, but as there are no code changes, compatibility should not be broken.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
             wordpress: 'nightly'
     steps:
       - name: Checkout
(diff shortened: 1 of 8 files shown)


stklcode:
Accepting plugin check failures here. Should be fixed, but none of the issues is new or critical, we want just a readme update for now.

PR #316, opened 2026-04-18
prepare release of v2.0.0-beta.1

stklcode:
* add CHANGELOG entries for 2.0.0
* add upgrade notice
* update screenshots


stklcode:
> Mismatched Stable Tag: 1.8.5 != 2.0.0-beta.1. Your Stable Tag is meant to be the stable version of your plugin and it needs to be exactly the same with the Version in your main plugin file's header. Any mismatch can prevent users from downloading the correct plugin files from WordPress.org.

That's OK here, beta.1 is not "stable". All other blocking checks now pass.

PR #323, opened 2026-05-17
prepare release of v2.0.0-beta.2

stklcode:
* update CHANGELOG entries for 2.0.0
* bump version

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -18,7 +18,7 @@
 /.stylelintrc.json
 /.sonarcloud.properties
 /composer.json
-/composer-php5.json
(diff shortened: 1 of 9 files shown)


stklcode:
> Error: Mismatched Stable Tag: 1.8.5 != 2.0.0-beta.2. Your Stable Tag is meant to be the stable version of your plugin and it needs to be exactly the same with the Version in your main plugin file's header. Any mismatch can prevent users from downloading the correct plugin files from WordPress.org.

Accepting this one for the **beta** release.

PR #329, opened 2026-06-01
Fix sprintf placeholder typo in referrers date range heading

thisismyurl:
Hey folks,

Spotted a small typo in the referrers heading string. When a date range is set, the format string uses %3$s twice — once in the "to" position and again for the post title — so the end date is silently dropped and the post title appears in both spots. The translator comment right above it correctly documents the intent (1: start date, 2: end date, 3: post title), so this is just a copy-paste slip in the string itself. One character fix.

(full disclosure, I used AI help with the testing and tweaks - Christopher)


stklcode (approved):
Hi Christopher.

Thanks for reporting and fixing this. Patch looks good.

PR #334, opened 2026-06-13
Move new menu item to sub menu

krafit:
I went for the path of least resistance and I'm not sure, this is my preferred solution. 

The alternative route would be one merged page. And registering a single slug and switch views via a `?view=` parameter inside `show_view()`. That would have been tidier, but it touches every internal `page=...` link in the three views and the `#statify_dashboard` selectors in `dashboard.js`.
I wanted to keep the edits light instead of doing a big restructuring of the codebase, but if we decide to go for the more complete approach, I'm game.


stklcode:
I like it. Though with same thoughs on the actual implementation.

Hidden menu pages are a pretty nice trick, but I think I'd prefer the "view" parameter here.

>That would have been tidier, but it touches every internal `page=...` link in the three views [...]

Yes, but it's not that many links. Maybe bundle the generation to a helper method like `dashboard_link( string $view )` if it gets too messy :thinking: But we have to take care about (hidden) form parameters, too.

> [...] and the `#statify_dashboard` selectors in `dashboard.js`. 

Really? We don't touch the DOM IDs within our views, so I guess there's no change required on this side.

`#statify_dashboard` is a selector for the dashboard widget anyway.


Sorry for taking over your branch, I've added a `view=...` refactoring als PoC, feel free to revert/drop the commit if we decide for another path.

krafit:
The more the merrier! The view approach would have been more or less the other way I contemplated, so no hard feelings at all.

inc/class-statify-evaluation.php
```diff
@@ -41,57 +41,87 @@ public static function add_capability(): void {
...
+	 * The evaluation is shown as a single "Statify" entry below the Dashboard
+	 * menu. The content and referrers pages are not registered as sub pages,
+	 * but accessible via in-page navigation.
 	 */
 	public static function add_menu(): void {
-		add_menu_page(
+		add_submenu_page(
+			'index.php',
 			__( 'Statify', 'statify' ),
 			'Statify',
```
stklcode: Why do we translate the page title but hard-code the menu title? 🤔

inc/class-statify-evaluation.php
```diff
@@ -41,57 +41,87 @@ public static function add_capability(): void {
...
+			return;
+		}
+
+		$last_key = array_key_last( $items );
+		?>
+		<nav class="statify-subnav wp-clearfix" aria-label="<?php echo esc_attr( $aria_label ); ?>">
+			<ul class="subsubsub">
+				<?php foreach ( $items as $key => $item ) : ?>
+				<li>
+					<a href="<?php echo esc_url( $item['url'] ); ?>"<?php echo empty( $item['current'] ) ? '' : ' class="current" aria-current="page"'; ?>><?php echo esc_html( $item['label'] ); ?></a><?php echo ( $key === $last_key ) ? '' : ' |'; ?>
```
stklcode: I would remove the separator from the markup and make it a design property, i.e. move it to CSS like this

```suggestion
					<a href="<?php echo esc_url( $item['url'] ); ?>"<?php echo empty( $item['current'] ) ? '' : ' class="current" aria-current="page"'; ?>><?php echo esc_html( $item['label'] ); ?></a>
```

```css
.statify-subnav li:not(:last-child)::after {
    content: "|";
    margin-left: 0.25em;
}
```

inc/class-statify-evaluation.php
krafit: whoops, that one slipped through.
Upon closer inspection we do the same thing in `class-statify-dashboard.php` 😅

stklcode: You're talking about the `wp_add_dashboard_widget()` call?

We have "Statify" translatable on the options menu and now for the evaluation menu. We could discuss whether it's a kind of brand name and not translatable or do it everywhere for consistency. It's implicitly translatable as part of various sentences, so I think I prefer the "everywhere" approach.
(not necessarily part of this PR)

=> #338

stklcode (approved):

PR #347, opened 2026-07-31
fix: sanitize and validate language tag in frontend l10n (#345)

### LINKED ISSUE #345: Broken frontend localization in 2.0.0
**Describe the bug**

Dashboards are broken, I.e. never finish loading, when a non-default user locale is selected.

```
Uncaught RangeError: invalid language tag: "en_GB" in dashboard.min.js
```

Apparently the notation is wrong, WP returns `en_GB` while Intl expects `en-GB` [^1]

**To Reproduce**

1. Change language to a non-default locale, like English (UK) or German (Austria)
2. Visit the dashboard or any Statify evaluation page

**Expected behavior**

Dashboards should render with localized date and number format.

- Version 2.0.0

**Additional context**

Originally reported in the forums: https://wordpress.org/support/topic/update-2-0-0-language-issue/

[^1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

stklcode:
WordPress i18n provides locales in POSIX notation like "en_GB", while JavaScript Intl expects BCP-47 like "en-GB". Wrong notation or unsupported languages trigger a RangeError that prevents the dashboard items to render properly.

Introduce logic to sanitize and validate the language tag wiht fallback to the user agent's language. This should fix the issue and provide a robust fallback, so a user may see the "wrong" notation in some edge cases, but should never experience an error.

----

resolves #345

### DIFF
--- a/js/dashboard.js
+++ b/js/dashboard.js
@@ -35,7 +35,7 @@
 	};
 
 	// Initialize number format.
-	const lang = wp.i18n.getLocaleData()['']?.lang || 'en';
+	const lang = getUserLanguage();
 	const numberFormat = new Intl.NumberFormat(lang);
 	const numberFormatPercent = new Intl.NumberFormat(lang, {
 		style: 'percent',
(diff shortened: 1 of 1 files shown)


js/dashboard.js
```diff
@@ -53,6 +53,25 @@
 	});
 	const dateFormatM = new Intl.DateTimeFormat(lang, { month: 'short' });
 
+	/**
+	 * Get user language from WP i18n, converted to BCP-47 notation.
+	 *
+	 * @return {string|undefined} BCP-47 language tag, if valid and supported.
+	 */
+	function getUserLanguage() {
+		let lc = wp.i18n.getLocaleData()['']?.lang || '';
```
2ndkauboy, line 62: **question**: Does the new code need no fallback to `en` anymore?

stklcode: May be discussed.

As the fallback comment below in the catch-block states, I changed the fallback value to `undefined` which means browser default if the value is unsupported (I’d keep that, even though `en` is likely never unsupported)

We could revert this first fallback to English. Unsure in which edge cases this might make a difference.

| value | 2.0.0 | proposal |
|---------|-------|----------|
| `en` | `en` | `en` |
| `en_GB` | error | `en-GB` |
| `xx` | undefined | undefined |
| `xx_XX` | error | undefined |
| `foo` | undefined | undefined |
| `en_FOO` | error | undefined |
| undefined | `en` | undefined |

PR #353, opened 2026-08-02
fix: add missing translations indicator for dashboard JS 

stklcode:
Frontend texts properly run through `wp.i18n`, but a backend call to `wp_set_script_translations()`[^1] was missing, so the translations never actually got loaded and JS-generated output contains defaults.

Originally reported in the forums: https://wordpress.org/support/topic/the-language-issue-in-the-statify-dashboard-widget/

[^1]: https://developer.wordpress.org/reference/functions/wp_set_script_translations/

### DIFF
--- a/inc/class-statify.php
+++ b/inc/class-statify.php
@@ -273,6 +273,7 @@ class Statify {
 				'sitename' => sanitize_key( get_bloginfo( 'name' ) ),
 			)
 		);
+		wp_set_script_translations( 'statify_chart_js', 'statify' );
 	}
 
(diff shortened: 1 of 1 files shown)


Zodiac1978:
Besides the already mentioned since and today I saw this in the tooltip too ("views" not translated).
Bildschirmfoto 2026-08-02 um 11 58 08

Patch fixes all three translation issues.

Zodiac1978 (approved):
Fix confirmed. No issues in the logs.
LGTM!

PR #356, opened 2026-08-02
Show html entities in JS

### LINKED ISSUE #355: HTML entities in the best list
**Describe the bug**
Using curly quotes or ndash/mdash in the title results in showing the HTML entities in the list:

Image

**To Reproduce**
1. Create a post with a curly quote, endash/mdash or another special character
2. Visit from incognito/private tab
3. Visit dashboard
4. See error in the best list

**Expected behavior**
Showing the character and not the Unicode like `&#8211;`

Zodiac1978:
Fixes #355

### DIFF
--- a/inc/class-statify-evaluation.php
+++ b/inc/class-statify-evaluation.php
@@ -468,7 +468,11 @@ class Statify_Evaluation extends Statify {
 			if ( 0 === $post_id ) {
 				$title = esc_url( $url );
 			} else {
-				$title = get_the_title( $post_id );
+				$title = html_entity_decode(
+					wp_strip_all_tags( get_the_title( $post_id ) ),
+					ENT_QUOTES | ENT_HTML5,
+					get_bloginfo( 'charset' )
+				);
 			}
 		}
 


inc/class-statify-evaluation.php
```diff
@@ -468,7 +468,11 @@ public static function post_title( ?string $url ): string {
 			if ( 0 === $post_id ) {
 				$title = esc_url( $url );
 			} else {
-				$title = get_the_title( $post_id );
+				$title = html_entity_decode(
+					wp_strip_all_tags( get_the_title( $post_id ) ),
```
stklcode, line 472: Does `wp_strip_all_tags()` do any real work here?

I'm not sure if there's any real-world scenario where the title can contain unencoded HTML tags. If that's a possibility, I'm fine with stripping it., otherwise we might simplify the expression.

Zodiac1978: If someone tries to add HTML via `the_title` filter. A safety guard. Not sure if this is _necessary_, but if it protects an attack vector and does not interfere with anything else ...

stklcode: Typing HTML into the title is encoded with/without stripping tags doesn't make any difference here:
Image

And that's exactly what the post title looks like in the blog.

stklcode: > If someone tries to add HTML via the_title filter.

OK, valid point. Doesn't really hurt and perfomance is not the strongest argument here, so let's strip it.

stklcode (approved):

PR #357, opened 2026-08-02
Change offset for chart not in the dashboard widget

### LINKED ISSUE #346: Description in diagram is cropped
**Describe the bug**
Image

**To Reproduce**
1. Go to Dashboard - Statify
2. Click on Content tab
3. See error

**Expected behavior**
No cropped description from the axes

**Additional context**
This happened on a 14" display, maybe this is relevant to the viewport width.

Zodiac1978:
Fixes #346

### DIFF
--- a/js/dashboard.js
+++ b/js/dashboard.js
@@ -482,6 +482,12 @@
 			return;
 		}
 
+		// Determine maximum value for scaling.
+		const maxValue = Math.max(...values);
+
+		// Generate dynamic offset roughly depending on the label length
+		const axisYOffset = Math.max(String(maxValue).length * 8, 24);
+
 		if (numericLabels) {
 			const container = root.parentElement;
 			let legend = container.querySelector('.statify-legend');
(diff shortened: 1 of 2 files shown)


js/dashboard.js
```diff
@@ -387,6 +387,9 @@
 		// Determine maximum value for scaling.
 		const maxValue = Math.max(...values);
 
+		// Use a wider Y axis on the extended Statify dashboard.
+		const axisYOffset = root.closest('.statify-chart-container') ? 64 : 30;
```
stklcode: 64 fits 8 digits including group markers (2x `[,.' ]` depending on the locale), but it's quite a lot of white space if numbers are lower like 1000 or less.

I tried using something dynamic like `String(maxValue).length * 8` which seems to work quite well (yields 64 for 8-figure numbers), probably with w lower bound like (`Math.min(24, ...)`) or maybe more elaborate with `numberFormat.format(maxValue).length`, but that may be little overdone.

And I would add this logic to the dashboard widget as well. 5-figure numbers per day with group separator are also cropped here at 30px.

stklcode: Dynamic with 8px per digit seems to work quite well. Lower bound at 24px, wondering if we may want to add an upper bound, too, maybe somewhere in the 80px region...

Image

Maybe better to go with suffixed notation like 1M 10M 100M if numbers get really large. But that can be optimized later, if necessary.

PR #360, opened 2026-08-07
prepare release of v2.0.2

stklcode:
* update version to 2.0.2
* update stable tag to 2.0.2
* add changelog entries
* add upgrade notice

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file. This projec
 
 ### Fixes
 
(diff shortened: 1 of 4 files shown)


Zodiac1978 (commented):
It looks like there's a wrong version number in the changelog.MD file the changelog entry shows 2.0.1 and not 2.0.2

stklcode:
Right ... :see_no_evil: 

At least the publicly visible changelog secion in _readme.txt_ is correct at 2.0.2

Issue #354, opened 2026-08-02
No tracking anymore since last update


stolperfrei-digital:
**Describe the bug**
Since Update to 2.0.1, not one single access to my website has been logged anymore

**To Reproduce**
1. Update to Statify 2.0.1
2. Use your website
3. no access is logged

**Expected behavior**
All visitors are logged like before

**System (please complete the following information):**
- WordPress 7.0.2
- Statify 2.0.1

stklcode:
Which tracking method do you use?

For JavaScript based tracking Statify 2.0 switched from using AJAX (calling endpoint `/wp-admin/admin-ajax.php`) to API (calling a custom endpoint `/wp-json/statify/v1/track`). This might lead to multiple issues in some non-standard scenarios:
* API access blocked for whatever reason.<br>Can be checked in browser console or server access logs looking for POST requests to the endpoint. HTTP 204 is expected behavior, 4xx errors indicate permissions issues.
* Old Statify JavaScript is still served from cache.<br>The script is included with version number as cache-breaker, but this may fall through some bundler/optimizer/caching plugins. Can also be checked in browser console, looking for _admin-ajax.php_ requests.

stklcode:
Assuming you are talking about the website that is named like your GitHub handle, I do see errors in my browser console:
The requests contains no nonce, so I assume "JS without nonce" is used. Trying to break site caches does not change the behavior.

```
POST https://***********.*******/wp-json/statify/v1/track 403 (Forbidden)
```

Best guess is some custom configuration blocks API access. I know there are some guides out there on the web that recommend such things (for reasons open to debate). If this is the case and you added some custom filters, those have to be adjusted for this specific endpoint to use Statify's built-in auth mechanism (configurable nonce-check, tracking is otherwise public by design).

Zodiac1978:
@stolperfrei-digital Do you use some snippet or plugin to restrict the REST-API? If yes, can you please share the snippet/plugin, so we can check what to do here? Thanks in advance!

stolperfrei-digital:
thanks. I take a look at the configuration. Yes I am using AIO WP Security. But that didn't change in the moment the tracking broke.

stolperfrei-digital:
Ok, I have to open the full REST-API in the settings as the statify endpoint is not listed in the list of whitlistable endpoints in the AIOS Plugin. So I cannot whitelist it. As the API was always protected, this worked before the last update.

Zodiac1978:
[...]

PR #364, opened 2026-08-20
fix: initialize REST API unconditionally

stklcode:
We initialize the API endpoints only on necessary routes and not on admin backend requests. Doing it unconditionally exposes the REST route to security plugins, so they can be enabled conveniently in restricted API configurations.

----

related to #354


Zodiac1978 (approved):
The change itself is fine and works as intended.

Should we document this change inline in some way?

stklcode:
Thought about that as well. Probably on the settings page in the description of the tracking method selector (JavScript tracking works via REST API, make sure the route "statify" route is not blocked.) And/or an FAQ section in the readme.

PR #366, opened 2026-08-24
fix(settings): localize built-in user role names on settings page

### LINKED ISSUE #361: User role names on settings page not translated
**Describe the bug**

On the _Statify_ settings page there's a list of user roles that should be allowed to see stats.
This list contains English names (Administrator, Editor, Author, Contributor, Subscriber) for built-in roles.

While custom roles might not be localized, the built-ins are, as we see in the user settings.

**To Reproduce**
1. Install Statify 2.x
2. Click on _Settings_ > _Statify_
3. Look at the user role selection (last option of widget settings)

**Expected behavior**
Role names should be loaclized - if possible.

- Version: 2.0.2
- WordPress 7.0.3

**Additional context**

Originally reported on the forums: https://wordpress.org/support/topic/missing-translation-in-the-settings/

faisalahammad:
## Summary

On the settings page the dashboard widget role list renders the built-in user role names in English even when the site runs in another language. Built-in roles (Administrator, Editor, Author, Contributor, Subscriber) are already translated by WordPress core for the user settings, but Statify outputs the raw stored names.

This change wraps the role display name with WordPress core's `translate_user_role()` so the built-in role names follow the active site language. Custom roles that are not present in the core translation files pass through unchanged, which matches the expected behavior in the issue.

### Before

```php
echo esc_html( $role_object['name'] );
```

### After

```php
echo esc_html( translate_user_role( $role_object['name'] ) );
```

Only the display name changes. The stored role slugs used as the checkbox value and id, the sanitization logic, and the widget visibility checks all keep the English slugs, so nothing about saving or permissions is affected.

### Why

`translate_user_role()` is a core WordPress function, available since WordPress 2.8, and uses the core text domain, so it picks up the translations WordPress already ships for the built-in roles. It is a no-op for custom roles not covered by those translations.

## Test steps

1. Install this branch in a WordPress site set to a non-English locale (for example German).
2. Go to Settings > Statify.
3. Check the dashboard widget role list (last widget option). The built-in role names should now appear translated.
4. Save the settings. Reopen the page. The selected roles are still checked, confirming the stored slugs are unchanged.
5. If a custom role is present, its name stays as defined.

## Screenshots

None. The change is display-only text localization.

Fixes #361

### DIFF
--- a/inc/class-statify-settings.php
+++ b/inc/class-statify-settings.php
@@ -305,7 +305,7 @@ class Statify_Settings {
 				esc_html( $name ),
 				checked( in_array( $role, $saved_roles, true ), true, false )
 			);
-			echo esc_html( $role_object['name'] );
+			echo esc_html( translate_user_role( $role_object['name'] ) );
 			?>
 			</label></p>
 			<?php
--- a/tests/test-settings.php
+++ b/tests/test-settings.php
@@ -190,8 +190,8 @@ class Test_Settings extends WP_UnitTestCase {
 			Statify::$options = $original_options;
 		}
 
-		self::assertStringContainsString(
-			'>Redakteur</label>',
+		self::assertMatchesRegularExpression(
+			'/>Redakteur\s*<\/label>/',
 			$html,
 			'built-in role display names should be localized via translate_user_role'
 		);
--- a/tests/test-settings.php
+++ b/tests/test-settings.php
@@ -139,4 +139,61 @@ class Test_Settings extends WP_UnitTestCase {
 			'unknown widget roles should have been removed'
 		);
 	}
+
+	/**
+	 * Test localization of built-in user role names in the roles list.
+	 */
+	public function test_show_roles_list_translates_names() {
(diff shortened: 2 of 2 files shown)


faisalahammad:
## CI Fix Summary — 1 failure resolved

| # | File | Error | Fix |
|---|------|-------|-----|
| 1 | tests/test-settings.php:193 | PHPUnit: expected output contains literally `>Redakteur</label>`, but template whitespace (tabs) sits between the name and `</label>` | Harden assertion to `assertMatchesRegularExpression('/>Redakteur\s*<\/label>/')`; translation via `translate_user_role()` was already correct in `inc/class-statify-settings.php` |

Tests ✅ (25/25, WP 6.2 · PHP 8.2 CI match) · Verification ✅ · PHP 7.4–8.5 compatible

stklcode (approved):
Hi @faisalahammad, thanks very much for the fix.

Works as expected, appreciate the additional tests.

PR #371, opened 2026-08-26
Right-align numbers in tables

patrickrobrecht:
Numeric columns in the statistics tables (dashboard widget, "Views"/"Proportion" columns on the content and referrer views) were left-aligned, which made columns of numbers harder to scan at a glance. This MR right-aligns them for better readability, consistent with common table conventions.

Some cells contained the `right` class already which was without effect, as the CSS did not contain a corresponding rule.

### DIFF
--- a/css/dashboard.css
+++ b/css/dashboard.css
@@ -86,6 +86,10 @@
 	width: 80%;
 }
 
+.statify-table .right {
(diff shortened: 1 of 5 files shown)


stklcode (approved):
Thanks, LGTM!

